PR #42, opened 2020-06-04
feat(askscript): function type constructor

mhagmajer:
`fun(returnType, arg1Type, arg2Type, ..., argNType)`

Last argument can be rest in this case last argument is `true`:

`fun(string, array(number), true)` - same as `(s: string, ...numbers: number[])` in TypeScript

Examples
```
const myToString: fun(string, int) = fun(n:int): string {
  n:toString
}
```

- [ ] update documentation


czerwinskilukasz1:
This PR changes function type used in .ask files to a new syntax. 
I think the new syntax is confusing, e.g. `fun(int, int, int)` doesn't look like the a *2*-argument function.
Let's chat about this change.

czerwinskilukasz1:
Also, this PR goes beyond the priorities from our Roadmap. 
Let's focus on our current priorities before we discuss this syntax change.

mhagmajer:
> This PR changes function type used in .ask files to a new syntax.
> I think the new syntax is confusing, e.g. `fun(int, int, int)` doesn't look like the a _2_-argument function.
> Let's chat about this change.

@czerwinskilukasz1 I'm all for creating some syntax sugar to make it easier to use, see #49

czerwinskilukasz1:
@mhagmajer , is this PR ready for review?  I'd say no, because it links issue #49 , which is about a syntax sugar `(int, int) -> int`, but there is not such syntax introduced in your change. Instead you use the confusing `fun(int, int, int)`.
I'd convert it to a draft. Thoughts?

mhagmajer:
@czerwinskilukasz1 this is not the recommended syntax for function type. This is an implementation detail - all types need constructors which are function that result in the given type. Function is a type of value so we need a function constructor for it. We can then have additional syntax sugar for it.

Because types are not checked by runtime yet, I created this PR only for the tests to showcase the basic syntax for creating function types.

Examples of function constructors:

```
const a: int({ positive: true }) = 5

const b: fun(int, int) = fun (a: int): int { 5 }
```

czerwinskilukasz1:
> Because types are not checked by runtime yet, I created this PR only for the tests to showcase the basic syntax for creating function types.
> 
> Examples of function constructors:
> 
> ```
> const a: int({ positive: true }) = 5
> 
> const b: fun(int, int) = fun (a: int): int { 5 }
> 
> // possible syntax sugar:
> const c: (int) -> int = fun (a: int): int { 5 }
> ```

The [ticket description](https://github.com/xFAANG/askql/issues/49) as well as [your description in this PR](https://github.com/xFAANG/askql/pull/42#issue-427773793) and our discussion in this PR were  about a different feature than you mentioned today.

mhagmajer:
@czerwinskilukasz1 can you explain please? I meant this PR to introduce the generic function type constructor

czerwinskilukasz1:
> @czerwinskilukasz1 can you explain please? I meant this PR to introduce the generic function type constructor

Of course. The linked issue is about `(int, int) -> int syntax`:
image

czerwinskilukasz1:
> Last argument can be rest in this case last argument is `true`:
> 
> `fun(string, array(number), true)` - same as `(s: string, ...numbers: number[])` in TypeScript

I find `fun(string, array(number), true)` syntax complex and confusing. How about we add `...` to AskScript too?

mhagmajer:
@czerwinskilukasz1 all syntax sugars are useful to make writing code easier :)

PR #76, opened 2020-06-07
fix(build): build and test now work on windows as well

YonatanKra:


config/postBuild.js
```diff
@@ -0,0 +1,4 @@
+const fs = require('fs');
+
+fs.copyFileSync('README.md', 'dist/README.md');
+console.log('README.md was copied to dist');
```
Tuhaj, line 4: missing new line

package.json
```diff
@@ -11,9 +11,9 @@
     "tsc": "tsc",
     "tsc:test": "tsc && npm run test --no-cache",
     "build:test": "npm run build && npm run test --no-cache",
-    "prebuild": "rm -rf dist && tsc",
+    "prebuild": "rimraf dist && tsc",
```
Tuhaj, line 14: yep, makes sense as on Windows you don't have `rm -rf`

config/postBuild.js
YonatanKra: Where? End of file?
Shouldn't prettier take care of that?

Tuhaj: Good point! Of course, it should!
I'm going to open the issue.

package.json
mhagmajer: Actually, let's use similar approach here and have a prebuild script which calls `rimraf`. Node LTS offers recursive removal too but its still experimental https://nodejs.org/dist/latest-v12.x/docs/api/fs.html#fs_fs_rmdir_path_options_callback

package.json
```diff
@@ -11,9 +11,9 @@
...
     "build": "jest --config jest.build.config.js --no-cache",
-    "postbuild": "cp README.md dist && cd dist && chmod +x *.js",
+    "postbuild": "node ./config/postBuild.js",
```
mhagmajer, line 16: Let's use TypeScript here and rename the scripts directory
```
"postbuild": "ts-node ./scripts/postBuild.ts",
```

config/postBuild.js
mhagmajer: We don't really use plain JavaScript in this repository so that's why prettier didn't do it. Also we don't run it automatically which actually is an issue

config/postBuild.js
```diff
@@ -0,0 +1,4 @@
+const fs = require('fs');
+
+fs.copyFileSync('README.md', 'dist/README.md');
```
mhagmajer, line 3: let's use the modern one here https://nodejs.org/dist/latest-v12.x/docs/api/fs.html#fs_fspromises_copyfile_src_dest_flags.

scripts/postBuild.ts
```diff
@@ -0,0 +1,5 @@
+import {promises} from 'fs';
```
mhagmajer, line 1: Please consider 
```ts
import { promises as fs } from 'fs';
```

or

```ts
import * as fs from 'fs/promises';
```

So that we can refer to these method as `fs` rather than `promises`

scripts/postBuild.ts
```diff
@@ -0,0 +1,5 @@
+import {promises} from 'fs';
+
+promises.copyFile('README.md', 'dist/README.md')
+    .then(() => console.log('README.md was copied to dist'))
+    .catch(err => { if (err) throw err;});
```
mhagmajer, line 5: In this project we prefer to use `async`/`await` syntax for dealing with promises. Please consider refactoring your code to the following

```ts
async main() {
  try {
    await fs.copyFile('README.md', 'dist/README.md');
    console.log('README.md was copied to dist');
  } catch (e) {
    console.error(e);
  }
}

main();
```

scripts/preBuild.ts
```diff
@@ -0,0 +1,3 @@
+import rimraf from 'rimraf';
+
+rimraf('./dist',error => { if (error) throw error;});
```
mhagmajer, line 3: async/await

```ts
async main() {
  try {
    await new Promise((resolve, reject) => {
      rimraf('./dist/', (error) => {
        if (error) {
          reject(error);
        } else {
          resolve();
        }
      });
    });
  } catch (e) {
    console.error(e);
  }
}

main();

PR #78, opened 2020-06-07
fix(askscript): update files after running tests and prettier

Tuhaj:


mhagmajer:
@Tuhaj are you sure that the right commit was included in this PR?

Issue #53, opened 2020-06-04
Generate .askcode and .askjsx in GitHub actions


mhagmajer:
Maybe GitHub action could also take care of generating `.askcode` and `.askjson` files? Currently these files are generated on `npm test`.  This way maybe they wouldn't have to be included in reviewed code. More reasearch is needed here on how this would work exactly.

mhagmajer:
Keeping this one open to track more things

mhagmajer:
Closing in favour of https://github.com/xFAANG/askql/issues/170

PR #80, opened 2020-06-08
feat(build): Run tests with GitHub actions

YonatanKra:
Solves #53

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,6 @@
+/dist
+/idea
+/node_modules
+/@types
+/testingP
+*-lock
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,6 +1,4 @@
 /dist
-/idea
+/.idea
 /node_modules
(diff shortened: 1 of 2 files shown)


.prettierignore
```diff
@@ -0,0 +1,6 @@
+/dist
+/idea
+/node_modules
+/@types
+/testingP
+*-lock
```
mhagmajer, line 6: seem like we don't have these at the top level

.prettierignore
```diff
@@ -0,0 +1,6 @@
+/dist
+/idea
```
mhagmajer, line 2: seem like we don't have these at the top level

YonatanKra: Correct :)

.prettierignore
YonatanKra: I've removed testingP and *-lock.
I do see @types when developing locally, so I guess it should also be ignored

PR #82, opened 2020-06-08
feat(CI): Prettier PR and push hooks

YonatanKra:
This PR will enforce a prettier check and commit on every push to a pull request.


.github/workflows/askql-ci.yaml
```diff
@@ -17,11 +17,17 @@ jobs:
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
       - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.head_ref }}
+      - uses: actions/setup-node@v1
+        with:
+          node-version: "12.x"
```
mhagmajer, line 24: do we need this?

.github/workflows/askql-ci.yaml
```diff
@@ -17,11 +17,17 @@ jobs:
...
+        with:
+          node-version: "12.x"
 
-      # Runs a single command using the runners shell
       - name: Checkout and install dependencies
         run: npm install
 
-      # Runs a set of commands using the runners shell
+      - name: Format
+        run: npm run format
```
mhagmajer, line 30: Can we make it so that this files if any formatting changes are required? The expectation is that the developer should send out correctly formatted files with this PR

PR #163, opened 2020-06-15
Update the demoServer example to use the askExpressMiddleware

YonatanKra:


mhagmajer:
A few nice ideas in there 🔥👏

PR #175, opened 2020-06-16
feat(askvm): implement pretty print for AskCode objects (closes #174)

### LINKED ISSUE #174: Pretty print AskCode
When debugging, reading long lines of AskCode programs is difficult.
Instead of reading
```
block(let('c',call(get('concat'),call(get('get'),'c'),' :)')),let('a',call(get('+'),call(get('get'),'a'),1)))
```
it would be much easier to read:
```
block(
  let(
    'c',
    call(
      get(
        'concat'
      ),
      call(
        get(
          'get'
        ),
        'c'
      ),
      ' :)'
    )
  ),
  let(
    'a',
    call(
      get(
        '+'
      ),
      call(
        get(
          'get'
        ),
        'a'
      ),
      1
    )
  )
)
```

Implement a pretty-print for AskCode objects.

czerwinskilukasz1:
closes #174


mhagmajer:
Pending discussion on https://github.com/xFAANG/askql/issues/174

src/askcode/lib/AskCode.ts
```diff
@@ -30,8 +30,8 @@ export class AskCode {
 
   parent?: AskCode;
 
-  toString(): string {
-    return askCodeToSource(this);
+  toString(pretty: boolean = false): string {
```
mhagmajer, line 33: let's take an option object here instead, so:

```ts
toString({
  pretty = false,
}: { pretty?: boolean } = {}): string {
```

This why it will be easier to develop this public API

czerwinskilukasz1: Right

czerwinskilukasz1:
> Let's keep `askCodeToSource` intact and use a separate function for pretty output of ask code.

Done

czerwinskilukasz1:
Thanks

Issue #182, opened 2020-06-17
Create an AskScript Playground from scratch


czerwinskilukasz1:
The current editor is a giant hack, we want have clean and well-structured code, hence use the editor.
Some notes on the editor can be found here: https://github.com/xFAANG/askql/issues/108

czerwinskilukasz1:
During my research it turned out that Monaco Editor does not support mobile devices. 
Also, Monaco Editor uses a completely different syntax highlighting than VS Code, so we'd need to implement it twice anyway, hence not much benefit from picking Monaco Editor.
Also, Playground should be playable on smartphones.

My pick is Ace editor, which is used in Cloud9, Amazon cloud-based IDE - so strong support guaranteed.
As for syntax highlighting for AskScript, good news too: "The Ace highlighting system is heavily inspired by the TextMate language grammar. ".
Moreover, they provide a tool which speeds up converting .tmlanguage file (used in VS Code) to their Ace highlighting system. This will likely mean much faster conversion.

mhagmajer:
> During my research it turned out that Monaco Editor does not support mobile devices.

I'm not concerned about people not being able to type code from their mobile devices.

> Also, Monaco Editor uses a completely different syntax highlighting than VS Code

Can you explain how exactly? I'm confused because it says on the Monaco Editor website that "The Monaco Editor is the code editor that powers VS Code". I would assume that the VS Code uses all its features including the code highlighting.

> Also, Playground should be playable on smartphones.

Sure. Playing save code examples should not require people to do any code editing.

czerwinskilukasz1:
>  I'm not concerned about people not being able to type code from their mobile devices.

I am. I think we shouldn't lock ourselves with an editor which doesn't support mobile if there are other editor doing a better job.

> Can you explain how exactly? I'm confused because it says on the Monaco Editor website that "The Monaco Editor is the code editor that powers VS Code". I would assume that the VS Code uses all its features including the code highlighting.

Could you trust me on this one?

mhagmajer:
> I am. I think we shouldn't lock ourselves with an editor which doesn't support mobile if there are other editor doing a better job.

Can you give some examples of programming language playgrounds that officially support editing code on mobile phones and tablets?

> Could you trust me on this one?

[...]

Issue #187, opened 2020-06-19
AskScript Playground: instead of sending AskScript send AskCode


czerwinskilukasz1:
Current Playground is sending AskScript to the server and server is parsing it to AskCode and then executing. 
What would show the idea of AskQL even better is to parse AskScript to AskCode in the browser and then send AskCode to the server, which is what would happen in real communication with an AskQL server. This requires adding AskScript parser to the Playground _frontend_.

czerwinskilukasz1:
Related to issue https://github.com/xFAANG/askql/issues/349

PR #189, opened 2020-06-19
feat(playground): add backend for AskScript Playground

czerwinskilukasz1:
It's a backend counter-part of #182 .


package.json
```diff
@@ -42,12 +42,16 @@
     "url": "https://github.com/xFAANG/askql/issues"
   },
   "homepage": "https://askql.org/",
-  "dependencies": {},
+  "dependencies": {
+    "cors": "^2.8.5",
+    "nanoid": "^3.1.10"
+  },
```
mhagmajer, line 48: These are all dev-dependencies for now. When we add package-level package.json file, then we can do this differentiation. This means that the playground is only available for development of this project.

src/demoAskScriptServer.ts
```diff
@@ -0,0 +1,154 @@
+#!/usr/bin/env node
```
mhagmajer, line 1: This should not be required to run this as script

package.json
czerwinskilukasz1: OK, done.

src/demoAskScriptServer.ts
czerwinskilukasz1: You added it to demoServer.ts, I just mimicked the same style.

image

Removing it from both files then.

czerwinskilukasz1:
> Same comments as in #87. Basically, what's different about this and demo server?
@mhagmajer 

Referring to your [comment here](https://github.com/xFAANG/askql/pull/87#discussion_r438127321) (having an odd endpoint that returns JavaScript code for adding HTML tags in browser):

The previous version accepted requests from the hacky frontend, so it needed to accept weird request structure and respond with even more weird response (a string with a piece of JS code, yuck).
This server returns a very clean response with the result data or an error message only.


Referring to [this comment](https://github.com/xFAANG/askql/pull/87#discussion_r438125917) (displaying logs):
I would like to have preview of what queries were run. It's very similar to the output you added to your server except that mine works better for multiple requests coming in at the same moment (this is what I used nanoid for).
We could use some sort of logging to file later on, but let's create the first version first and then work further to make it perfect. 

Referring to [this comment](https://github.com/xFAANG/askql/pull/87#discussion_r438125831) (working the same as demoServer does):
The server currently accepts AskScript not AskCode.  The communication model we promote is using AskCode for network communication, so modifying a great example of an AskCode server doesn't sound right to me. We should keep the ultimate version of the server clean and not contaminate it with such intermittent changes.

Referring to [this comment](https://github.com/xFAANG/askql/pull/87#discussion_r438124257) (having an endpoint for /, which redirects a person to askql.org website):
Because why not? I wanted to use /askscript endpoint for accepting the code to underline it's AskScript not AskCode sent over (will change in the future, we talked about this).

Does it answer your questions?

mhagmajer:
> I would like to have preview of what queries were run.

How are you going to capture the backend `console.log` output and use it for diagnosticts?

> The server currently accepts AskScript not AskCode. The communication model we promote is using AskCode for network communication, so modifying a great example of an AskCode server doesn't sound right to me. We should keep the ultimate version of the server clean and not contaminate it with such intermittent changes.

Both the demoServer and server from this PR accept AskScript as input. If you wish to add an option for the program to come in as AskCode, you can achieve that for example by adding option for it with a URL param. I disagree on forking the same feature because we want a different option to it. Let's reuse the what we can.

> Referring to this comment (having an endpoint for /, which redirects a person to askql.org website):
Because why not? I wanted to use /askscript endpoint for accepting the code to underline it's AskScript not AskCode sent over (will change in the future, we talked about this).

Demo server can have whatever endpoints we want. Let's make it clear what they are and how people should use them

czerwinskilukasz1:
> > I would like to have preview of what queries were run.
> 
> How are you going to capture the backend `console.log` output and use it for diagnosticts?

Using `tee` or a redirect to a log file works well as a first version. I'm already using it so. 
There are of course hundreds of ways to log things, there also dozens of other aspects of the code we could discuss now. However, how about we focus on making progress in iterations instead of trying to work on a perfect solution in one commit? 

> > The server currently accepts AskScript not AskCode. The communication model we promote is using AskCode for network communication, so modifying a great example of an AskCode server doesn't sound right to me. We should keep the ultimate version of the server clean and not contaminate it with such intermittent changes.
> 
> Both the demoServer and server from this PR accept AskScript as input. If you wish to add an option for the program to come in as AskCode, you can achieve that for example by adding option for it with a URL param. I disagree on forking the same feature because we want a different option to it. Let's reuse the what we can.

I run the server from master 2 days ago and it accepted AskCode but threw errors on any AskScript. 
Now I see the source code which uses `askScriptCode` variable but in fact it is AskCode. 
https://github.com/xFAANG/askql/blame/master/src/askExpressMiddleware/askExpressMiddleware.ts#L20
A bug? Filing it here https://github.com/xFAANG/askql/issues/191 

> 
> > Referring to this comment (having an endpoint for /, which redirects a person to askql.org website):
> > Because why not? I wanted to use /askscript endpoint for accepting the code to underline it's AskScript not AskCode sent over (will change in the future, we talked about this).
> 
> Demo server can have whatever endpoints we want. Let's make it clear what they are and how people should use them

Ok, if so, I will use the demoServer.ts when the bug gets fixed.

mhagmajer:
> Using tee or a redirect to a log file works well as a first version.

Just wondering how this is going to work in deployment. Let's say we deploy this to Netlify Function. How are we going to access these logs?

Thanks for filing the bug for express middleware, let's discuss it in the ticket

czerwinskilukasz1:
> > Using tee or a redirect to a log file works well as a first version.
> 
> Just wondering how this is going to work in deployment. Let's say we deploy this to Netlify Function. How are we going to access these logs?

In which milestone would you like to deploy it to Netlify?

src/demoAskScriptServer.ts
```diff
@@ -0,0 +1,152 @@
+import bodyParser from 'body-parser';
```
mhagmajer, line 1: as I understand we won't need to add this once https://github.com/xFAANG/askql/issues/191 is resolved, correct?

src/demoAskScriptServer.ts
```diff
@@ -0,0 +1,152 @@
...
+app.use(cors());
+app.use(bodyParser.json());
+app.use(bodyParser.urlencoded({ extended: true }));
+
+app.get('/', async (req, res) => {
+  res.type('html');
+  res.send(
+    '<html><head><title>AskQL</title></head><body><p><b>Ask</b> me anything!</p><p><a href="http://askql.org/">AskQL website</a></p></body></html>'
+  );
+});
```
mhagmajer, line 89: if this is to serve as some human-readable documentation for the server it'd be good to extend it a little bit and maybe list those custom resources very simply?

src/demoAskScriptServer.ts
```diff
@@ -0,0 +1,152 @@
...
+
+app.post('/askscript', async (req, res) => {
```
mhagmajer, line 91: Since we're probably going to support different options for providing the query, let's maybe use a more generic endpoint name - I suggest `/ask`

package.json
```diff
@@ -27,7 +27,7 @@
     }
   },
   "lint-staged": {
-    "*": [
+    "*.{js,jsx,ts,tsx}": [
```
mhagmajer, line 30: Can you explain what this does and why you decided to add this in this PR?

mhagmajer:
> In which milestone would you like to deploy it to Netlify?

I used Netlify as an example. I assume this playground backend is going to be deployed to a server. I am wondering how are we going to get these logs from that server.

package.json
czerwinskilukasz1: I couldn't add files to my commit because prettier & eslint were running against unsupported file types and complained about errors.

src/demoAskScriptServer.ts
czerwinskilukasz1: The resources are provided in Playground frontend as a link to the source code.

src/demoAskScriptServer.ts
czerwinskilukasz1: Correct.  We can easily merge this PR and then later replace this file with the express code when it works.

src/demoAskScriptServer.ts
czerwinskilukasz1: > > In which milestone would you like to deploy it to Netlify?
> 
> I used Netlify as an example. I assume this playground backend is going to be deployed to a server. I am wondering how are we going to get these logs from that server.
> 
> @czerwinskilukasz1 Łukasz, here's an example solution for logging [http://github.com/winstonjs/winston](https://github.com/winstonjs/winston), there are more

Great, let's discuss it in a separate issue. https://github.com/xFAANG/askql/issues/215

czerwinskilukasz1:
> > In which milestone would you like to deploy it to Netlify?
> 
> I used Netlify as an example. I assume this playground backend is going to be deployed to a server. I am wondering how are we going to get these logs from that server.
> 
> @czerwinskilukasz1 Łukasz, here's an example solution for logging [http://github.com/winstonjs/winston](https://github.com/winstonjs/winston), there are more

Great, let's discuss it in a separate issue. https://github.com/xFAANG/askql/issues/215

package.json
mhagmajer: Let's solve this issue with the configuration of these respective tools. Maybe the way we run prettier in the CI doesn't read `prettier.config.js`? If you want to go ahead with this quick fix, could you also create a bug issue to fix it in the right place?

src/demoServer.ts
```diff
@@ -66,7 +64,7 @@ app.use(bodyParser.urlencoded({ extended: true }));
 
 app.post('/ask', [logRequestMiddleware, askMiddleware, logResultsMiddleware]);
 
-const port = 80;
+const port = 1234;
```
mhagmajer, line 67: maybe let's use env var for this as well?

src/demoAskScriptServer.ts
mhagmajer: Can you create a issue for it please so we know that this needs a follow up?

mhagmajer: Seems like we can do on it already

package.json
czerwinskilukasz1: Great, let's do so. 
https://github.com/xFAANG/askql/issues/221

src/demoAskScriptServer.ts
czerwinskilukasz1: > Please use @YonatanKra's change

@mhagmajer , I think Yonatan's server is not ready to be used in AskScript Playground. It assumes the code is in AskCode and only if parsing it fails, it tries to parse it as an AskScript program. Therefore, it cannot be used as a server for AskScript Playground, where user is expecting to write only valid AskScript programs.

Example:
Typing in the AskScript Playground:
`ask(let('plus',fun(let('a',get('$0')),let('b',get('$1')),let('c',get('$2')),return(call(get('sum'),call(get('sum'),get('a'),get('b')),get('c'))))),call(get('plus'),2,3.6,4))` should result in an error but if I used Yonatan's server, user would see result: `9.6`

czerwinskilukasz1:
> Please use @YonatanKra's change

@mhagmajer , I think Yonatan's server is not ready to be used in AskScript Playground. It assumes the code is in AskCode and only if parsing it fails, it tries to parse it as an AskScript program. Therefore, it cannot be used as a server for AskScript Playground, where user is expecting to write only valid AskScript programs.

Example:
Typing in the AskScript Playground:
`ask(let('plus',fun(let('a',get('$0')),let('b',get('$1')),let('c',get('$2')),return(call(get('sum'),call(get('sum'),get('a'),get('b')),get('c'))))),call(get('plus'),2,3.6,4))` should result in an error but if I used Yonatan's server, user would see result: `9.6`

src/demoServer.ts
czerwinskilukasz1: I shouldn't have committed this change. Removed.

src/demoAskScriptServer.ts
czerwinskilukasz1: Let's keep it /askscript for now, because it signals that Ask*Script* is accepted not AskCode. I had created an issue for sending AskCode to the server instead of AskScript. We could change the name of the endpoint then (or introduce /ask in addition to /askscript), let's keep the conversation here: https://github.com/xFAANG/askql/issues/187

src/demoAskScriptServer.ts
czerwinskilukasz1: Alas, not yet.  https://github.com/xFAANG/askql/pull/189#issuecomment-648822434

YonatanKra:
> > Please use @YonatanKra's change
> 
> @mhagmajer , I think Yonatan's server is not ready to be used in AskScript Playground. It assumes the code is in AskCode and only if parsing it fails, it tries to parse it as an AskScript program. Therefore, it cannot be used as a server for AskScript Playground, where user is expecting to write only valid AskScript programs.
> 
> Example:
> Typing in the AskScript Playground:
> `ask(let('plus',fun(let('a',get('$0')),let('b',get('$1')),let('c',get('$2')),return(call(get('sum'),call(get('sum'),get('a'),get('b')),get('c'))))),call(get('plus'),2,3.6,4))` should result in an error but if I used Yonatan's server, user would see result: `9.6`

@czerwinskilukasz1 
I do not understand the issue. Why should this code result in an error?

mhagmajer:
That is actually an interesting point that since AskCode and AskScript have separate syntax trees we could have a single parser in playground that supports both of these.

czerwinskilukasz1:
Thanks, Marcin

PR #196, opened 2020-06-19
feat(playground): add frontend for AskScript Playground

czerwinskilukasz1:
The new Playground frontend linked to the new Playground backend is available at:
http://cli.askql.org:5000/

image

Some design polish from Kari is coming soon.


package.json
```diff
@@ -16,7 +16,7 @@
     "build": "jest --config jest.build.config.js --no-cache",
     "postbuild": "ts-node scripts/postBuild.ts",
     "prettier": "prettier --write --config prettier.config.js .",
-    "prettier:check": "prettier --config prettier.config.js --check .",
+    "prettier:check": "prettier --config prettier.config.js --check '**/*.{js,jsx,ts,tsx}'",
```
mhagmajer, line 19: please explain why we need that

src/playground/dist/app.js
```diff
@@ -0,0 +1,53 @@
+'use strict';
```
mhagmajer, line 1: This looks like compiled version of [src/playground/src/app.ts ](https://github.com/xFAANG/askql/pull/196/files#diff-231aef61cb7355918bb7bcaee3989cff)

src/playground/copyStaticAssets.ts
```diff
@@ -0,0 +1,6 @@
+import * as shell from 'shelljs';
+
+shell.mkdir('-p', 'dist/public/');
+shell.cp('-R', 'src/public/assets/css', 'dist/public/');
+shell.cp('-R', 'src/public/assets/js', 'dist/public/');
+shell.cp('-R', 'src/views', 'dist/');
```
mhagmajer, line 6: We've been using the native Node `fs` library for this sort of thing - https://github.com/xFAANG/askql/blob/master/scripts/postBuild.ts. Do you think it's worth it to add this additional dependency?

src/playground/dist/public/js/app.js
```diff
@@ -0,0 +1,66 @@
+function registerAskScriptEditor(
```
mhagmajer, line 1: Would it be possible to do this in TypeScript?

src/playground/package.json
```diff
@@ -0,0 +1,64 @@
+{
+  "name": "askscript-playground",
+  "version": "1.0.0",
+  "main": "index.js",
+  "scripts": {
+    "build": "mkdir -p dist/public && npm run build-assets && npm run build-ts && npm run copy-static-assets",
```
mhagmajer, line 6: How is this going to integrate with the build process of the entire repo?

src/playground/dist/views/main.handlebars
```diff
@@ -0,0 +1,88 @@
+<!doctype html>
```
mhagmajer, line 1: Why did you pick Handlebars here instead of React or [Hybrids](https://github.com/hybridsjs/hybrids)?

src/playground/dist/public/js/app.js
```diff
@@ -0,0 +1,66 @@
...
+    }
+  } catch (e) {
+    showErrorResponse('Got an error when parsing the response');
+  }
+}
+
+function showSuccessfulResponse(result) {
+  const resultElem = document.getElementById('result');
+  resultElem.innerText = JSON.stringify(result, null, 2);
+  resultElem.classList.remove('error');
```
mhagmajer, line 59: Due to unwieldiness of the DOM it's difficult to maintain code which deals with this API directly

src/playground/dist/public/js/app.js
mhagmajer: Also, there seems to be a copy of this file in `src/playground/src/public/assets/js/app.js`

km4:
@czerwinskilukasz1  Why we push the dist folder on repo? It should be generated by build process.

src/playground/dist/app.js
czerwinskilukasz1: You're right, I didn't notice including `dist/` folder. Removing it now.

src/playground/copyStaticAssets.ts
czerwinskilukasz1: Yes, becase surprisingly, there is no function to copy the whole directory in `fs`.
https://nodejs.org/api/fs.html

Please use their own functions (which may contain bugs) or one of libraries.
E.g. these: https://stackoverflow.com/questions/13786160/copy-folder-recursively-in-node-js

src/playground/package.json
czerwinskilukasz1: Not sure what you mean. 
monorepo means many packages in one repository, so that code can be shared between them. 
Every package has their own package.json file with their own build instructions. Therefore, there is no single build process as the packages are independent, same as they would be if they were in separate repositories.

src/playground/dist/views/main.handlebars
czerwinskilukasz1: Because it's minimal. In fact, we could live even without any templates at all in this project, so I chose the most minimalistic solution.

src/playground/package.json
mhagmajer: I'm asking because we don't currently have package-level package.json. Why introduce it here? Is build process of this package different than that of AskQL? How so? If it was the same, then we wouldn't need to add this file. What about releases of the playground? Is this going to NPM?

src/playground/src/public/assets/js/app.js
```diff
@@ -0,0 +1,66 @@
...
+    }
+  } catch (e) {
+    showErrorResponse('Got an error when parsing the response');
+  }
+}
+
+function showSuccessfulResponse(result) {
+  const resultElem = document.getElementById('result');
+  resultElem.innerText = JSON.stringify(result, null, 2);
+  resultElem.classList.remove('error');
```
czerwinskilukasz1, line 59: Following [your comment](https://github.com/xFAANG/askql/pull/196#discussion_r443121682) here:

> Due to unwieldiness of the DOM it's difficult to maintain code which deals with this API directly

What do you mean by 'unwieldiness of the DOM'?  [MDN is saying](https://developer.mozilla.org/en-US/docs/Web/API/Element/classList) this is a good way to go, supported by all major browsers.
What's your proposal?

czerwinskilukasz1:
> @czerwinskilukasz1 Why we push the dist folder on repo? It should be generated by build process.

By mistake. Fixed. Thanks.

src/playground/copyStaticAssets.ts
mhagmajer: Ok, cool. However, I think this should be a part of https://github.com/xFAANG/askql/tree/master/scripts

src/playground/package.json
czerwinskilukasz1: Marcin, Playground frontend is a separate project. It's separate from AskVM and AskScript parser both in terms of a separate codebase and in terms of dependencies.
It has a number of dependencies not needed in any other project, like `imagemin-cli, node-sass, webpack, webpack-cli, @types/express-handlebars` to name a few.
Adding them to the main package.json would create a lot of clutter and a mix of dependencies needed by AskVM and Playground.

src/playground/src/public/assets/js/app.js
mhagmajer: Well, I would use React here.

https://softwareengineering.stackexchange.com/questions/147451/whats-so-bad-about-the-dom

mhagmajer:
Can we use TypeScript instead of JavaScript?

package.json
czerwinskilukasz1: Due to a bug in Prettier, when running `--check` Prettier checks all the files even non-supported ones.  This caused the build step on Github to fail. Hence my change.

src/playground/src/public/assets/js/app.js
czerwinskilukasz1: Why would you introduce a heavy dependency in such a lightweight project just for one line of code?

czerwinskilukasz1:
> Can we use TypeScript instead of JavaScript?

We could.

czerwinskilukasz1:
> Good progress. Here's some remaining points that I think need to be addressed.
> 
> * [ ]  remove local package.json - let's reuse the current build process and TS config for simplicity and consistency
> * [ ]  remove local .gitignore, LICENSE (it's the same),

@mhagmajer, I addressed this [here](https://github.com/xFAANG/askql/pull/196#discussion_r444174380) and don't see your comment.

> * [ ]  rename local README to README.md
ok

> * [ ]  specify in the UI what version of AskQL this runs
ok

 
> Follow-up things than can be done later - please create issues
> 
> * [ ]  migrated from Handlebars to React

https://github.com/xFAANG/askql/issues/216

> * [ ]  support running askvm on client side - doesn't require server

https://github.com/xFAANG/askql/issues/217

mhagmajer:
> Marcin, Playground frontend is a separate project. It's separate from AskVM and AskScript parser both in terms of a separate codebase and in terms of dependencies.
> It has a number of dependencies not needed in any other project, like `imagemin-cli, node-sass, webpack, webpack-cli, @types/express-handlebars` to name a few.
> Adding them to the main package.json would create a lot of clutter and a mix of dependencies needed by AskVM and Playground.

I'm not concerned about mixing this as most of these are development time dependencies. The important dependencies are the direct dependencies - we don't have any of those. It's important that we keep single package.json file so that all script to build and deploy are in one place. Can you add commands needed to build the playground there, please?

Also, no need to have a separate configuration for TypeScript and a separate compilation step. This is still a small project - we can compile all code in the same command

package.json
mhagmajer: Commenting for reference: https://github.com/xFAANG/askql/issues/221

.env.example
```diff
@@ -0,0 +1,9 @@
+
+# Note we depend on NODE_ENV being set to dictate which of the env variables below get loaded at runtime. 
+
+# Application Port - express server listens on this port (default 3000).
+PLAYGROUND_PORT=3000
+
+# URL of the AskScript backend, including path
+PLAYGROUND_ASK_SERVER_URL="http://cli.askql.org:6001/askscript"
+PLAYGROUND_ASK_SERVER_VERSION_URL="http://cli.askql.org:6001/version"
```
mhagmajer, line 9: Maybe it would be easier to just have url without path? We might want more flexibility with endpoints

mhagmajer: How is the port configured for askql server? Why is it `6001` here if "PLAYGROUND_PORT" is `3000`?

README.md
```diff
@@ -75,18 +75,22 @@ Please find all important information here:
 
 1. Clone the repository
 
-`git clone git@github.com:xFAANG/askql.git`
+   `git clone git@github.com:xFAANG/askql.git`
 
-2. Install dependencies:
+2. It is advised to install autoreconf as it is used by one of the Node packages used by AskScript Playground.
```
mhagmajer, line 80: What is autoreconf? Does it apply only to Linux? If so maybe, maybe we can explain that in the beginning or make this step applicable only for Linux development

README.md
```diff
@@ -242,6 +246,43 @@ string ask(call(get('find'),get('philosophers'),fun(let('name',get('$0')),call(g
 
 Here is the link to our [AskQL playground](http://cli.askql.org:3000/)!
 
+#### Developer info - how to run
+
+1. Copy .env.example to .env and set `PLAYGROUND_PORT`, `PLAYGROUND_ASK_SERVER_URL` and `PLAYGROUND_ASK_SERVER_VERSION_URL` appropriately.
```
mhagmajer, line 251: Changing this values shouldn't be required by default

package.json
```diff
@@ -13,13 +13,24 @@
...
     "prettier:scripts": "prettier --write ./scripts",
     "lint:js": "eslint . --ext .js,.tsx,.ts",
-    "lint:js:fix": "npm run lint:js --fix"
+    "lint:js:fix": "npm run lint:js --fix",
+    "playground:build": "rm -rf dist/playground && mkdir -p dist/playground/public && npm run playground:build-assets && npm run playground:build-ts && npm run playground:copy-static-assets",
+    "playground:build-assets": "npm run playground:build-assets:img",
+    "playground:build-assets:img": "imagemin src/playground/public/assets/img/* --out-dir=dist/playground/public/img",
+    "playground:build-ts": "tsc && mv dist/playground/public/assets/js dist/playground/public/js",
+    "playground:copy-static-assets": "ts-node src/playground/copyStaticAssets.ts",
+    "playground:start": "node dist/playground/app.js",
```
mhagmajer, line 28: `ts-node src/playground/app.ts` ?

package.json
```diff
@@ -13,13 +13,24 @@
...
+    "playground:watch-dev": "npm-watch playground:start-dev",
```
mhagmajer, line 29: `ts-node-dev src/playground/app.ts` ?

I don't know the `npm-watch`, however, we already use `ts-node`

package.json
```diff
@@ -13,13 +13,24 @@
...
+    "playground:start-dev": "node dist/playground/app.js"
+  },
+  "watch": {
+    "playground:start-dev": "dist/public/css/*.css"
```
mhagmajer, line 33: So this only watches build target? 🤔

package.json
```diff
@@ -48,14 +59,31 @@
...
     "node": ">=12.14.0"
   },
-  "dependencies": {},
+  "dependencies": {
+    "dotenv": "^8.2.0",
+    "express": "^4.17.1",
+    "express-handlebars": "^4.0.4",
+    "handlebars": "^4.7.6",
+    "shelljs": "^0.8.4"
+  },
```
mhagmajer, line 68: Playground only needs to work for development, not in a shipped package. Therefore, these are all development dependencies. If someone wants to deploy the playground they can always do `npm install --only=dev`

package.json
```diff
@@ -48,14 +59,31 @@
...
   "devDependencies": {
+    "@babel/polyfill": "^7.10.1",
+    "@types/ace": "0.0.43",
+    "babel-loader": "^8.1.0",
+    "cz-conventional-changelog": "^3.2.0",
```
mhagmajer, line 73: ?

package.json
```diff
@@ -88,7 +119,13 @@
...
     "prettier": "2.0.5",
     "rimraf": "3.0.2",
+    "shelljs": "^0.8.4",
     "ts-node": "8.10.2",
     "typescript": "3.9.5"
+  },
+  "config": {
+    "commitizen": {
+      "path": "./node_modules/cz-conventional-changelog"
+    }
```
mhagmajer, line 129: what is this?

src/playground/README
```diff
@@ -0,0 +1,34 @@
+## How to install
+
+It is advised to install autoreconf as it is used when installing one of the Node packages. For Ubuntu/Debian run:
+
+    sudo apt-get install autoconf
```
mhagmajer, line 5: It's not clear to what OSs this applies

src/playground/README
```diff
@@ -0,0 +1,34 @@
...
+
+Install Node packages:
+
+    npm install
+
+
+## How to run
+
+1. Copy .env.example to .env and edit PORT and ASKSCRIPT_SERVER_URL.
```
mhagmajer, line 14: Is editing necessary?

src/playground/README
```diff
@@ -0,0 +1,34 @@
+## How to install
```
mhagmajer, line 1: Maybe the main readme, can just link here?

src/playground/app.ts
```diff
@@ -0,0 +1,53 @@
...
+import dotenv from 'dotenv';
+import fs from 'fs';
+
+if (fs.existsSync('.env')) {
+  console.log('Using .env file to supply config environment variables');
+  dotenv.config({ path: '.env' });
+} else {
+  console.log('Using .env.example file to supply config environment variables');
+  dotenv.config({ path: '.env.example' }); // you can delete this after you create your own .env file!
+}
```
mhagmajer, line 14: https://www.npmjs.com/package/dotenv-cli does this for you. Let's not read from `.env.example` we could copy `.env.example` to `.env` if the latter doesn't exist in the `postinstall` phase

src/playground/app.ts
```diff
@@ -0,0 +1,53 @@
...
+
+const app = express();
+
+app.use(
+  express.static(path.join(__dirname, 'public'), { maxAge: 31557600000 })
```
mhagmajer, line 19: `Infinity` will do as good as 1000 years :D

How does the caching work if someone updates assets in `/public` for development?

src/playground/app.ts
```diff
@@ -0,0 +1,53 @@
...
+// handlebars
+app.engine(
+  'handlebars',
+  expressHandlebars({
+    layoutsDir: __dirname + '/views',
+    defaultLayout: 'main',
+  })
+);
+app.set('views', __dirname + '/views');
+app.set('view engine', 'handlebars');
```
mhagmajer, line 31: Do we need this? Seems like this information has been already provided to `app.engine`

src/playground/views/main.handlebars
```diff
@@ -0,0 +1,215 @@
...
+	{{#if gtm}}
+	<!-- Global site tag (gtag.js) - Google Analytics -->
+	<script async src="https://www.googletagmanager.com/gtag/js?id={{gtm}}"></script>
+	<script>
+		window.dataLayer = window.dataLayer || [];
+		function gtag() { dataLayer.push(arguments); }
+		gtag('js', new Date());
+		gtag('config', '{{gtm}}');
+	</script>
+	{{/if}}
```
mhagmajer, line 18: Not sure if adding tracking code in a public repository is okay... This would mean that we need to get users consent for storing their tracking data

czerwinskilukasz1: @Tuhaj , what do you think?

src/playground/README
czerwinskilukasz1: As the line is saying, the command applies to Ubuntu/Debian.

src/playground/app.ts
czerwinskilukasz1: It will cache until, well, max-age I think. 
This is why 0 seems a better value.

src/playground/README
czerwinskilukasz1: I figured out that this README would be the only readme file in subdirectories, so I am removing it, same as I removed package.json and LICENSE. Let's keep things in one place if we have a monorepo.  
All the contents are in the main README.md anyway.

package.json
czerwinskilukasz1: Removed  cz-conventional-changelog

package.json
czerwinskilukasz1: ok.

package.json
czerwinskilukasz1: Thanks. These dev commands are not working, removing.

package.json
czerwinskilukasz1: removed

src/playground/app.ts
czerwinskilukasz1: Seems so: https://www.npmjs.com/package/express-handlebars has both app.engine and app.set('view engine')

package.json
czerwinskilukasz1: Playground should use dist/, same as the rest of the project.

README.md
czerwinskilukasz1: It's impossible to guess what URL for the backend the user would use.
The default will work with our Playground backend on cli.ask.org after I replace the old Playground with the new one.

README.md
czerwinskilukasz1: The next line is stating Ubuntu/Debian.  What else would you add here?

package.json
czerwinskilukasz1: These dev commands are not working, removing.

src/playground/app.ts
czerwinskilukasz1: Done. 
Note that I needed to create a separate postBuild.ts file so that `npm run playground:build` works properly.

README.md
km4: if you skip this step and run the server then the server gets variables from the env.example.

src/playground/views/main.handlebars
czerwinskilukasz1: Moved to a separate https://github.com/xFAANG/askql/issues/235

src/playground/copyStaticAssets.ts
czerwinskilukasz1: Done

czerwinskilukasz1:
> Let's move the binary files to CDN (simply we could host them in the same place where we host askql.org) to keep the package size small https://bundlephobia.com/result?p=askql@1.1.0
> 
> A lot of scripts that this playground depends on is already loaded from CDN

If we keep adding new requirements and enhancements, we will never publish this PR. Let's discuss this enhancement in a separate issue: https://github.com/xFAANG/askql/issues/237


> Also, seems like the discussion Ace vs Monaco hasn't concluded yet #182

Let's keep the PR smaller and discuss this topic separately: https://github.com/xFAANG/askql/issues/236

README.md
mhagmajer: It's not clear to me if I need to think about `autoreconf` if I'm on MacOS or Windows

src/playground/public/assets/js/demos.ts
```diff
@@ -0,0 +1,25 @@
...
+    const a = document.createElement('a');
+    a.setAttribute('class', 'dropdown-item');
+    a.setAttribute('href', '#');
+    a.onclick = (e) => {
+      e.preventDefault();
+      pickDemo(editor, key);
+    };
+    a.innerText = demos[key].title;
+    menu.appendChild(a);
+  }
```
mhagmajer, line 15: Not sure why you're not using Handlebars for this https://handlebarsjs.com/guide/builtin-helpers.html#each... Also, templates for frontend are quite old-fashioned

src/playground/public/assets/js/demosData.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  demo20: {
+    title: 'Factorial',
+    code: `ask {
+  const factorial: int(int) = fun(n:int):int {
+    if (n:lessThan(2)) {
+      return n
+    }
+
+    n:multiply(factorial(n:minus(1)))
```
mhagmajer, line 23: probably operators would make easier code here

czerwinskilukasz1:
Thanks

Issue #170, opened 2020-06-16
Don't commit .askjson and .askcode anymore


czerwinskilukasz1:
Following a discussion on [Discord channel](https://discord.com/channels/688128418914500838/704117328735895662/722427091210993714), eventually, once we get to certain level of stability, we could replace .askjson and .askcode with some unit tests and stop commit these files.

This is a task for the future, several things need to happen before we consider doing it.

YonatanKra:
I do not believe these files need to be committed - they can be just placed in `.gitignore`. They are being auto-generated on every test run anyway...
What do you think? Would that be a good solution?

mhagmajer:
@YonatanKra @czerwinskilukasz1 let's come up with a long .ask file that uses all of our constructs and lets commit snapshots only for this file. This way we keep most value of snapshots while decluttering PRs. Who's up for it?

czerwinskilukasz1:
Tasks:

- [ ] Create one example per AskScript construct in `src/askscript/__tests__/00-documentation-examples/` and add them to Documentation on Notion
- [ ] Stop creating .snapshot.jsx and .askcode for tests in other folders (configured in a config file)

PR #199, opened 2020-06-20
fix: variable scopes with let and assign

### LINKED ISSUE #129: Assigned value is discarded when exiting an inner scope
```
ask {
    let a = 2

    let c = 'Hello'
    
    if(a < 5) {
        c = '!'
    }
    
    return c
}
```
returns `Hello`
instead of `!`

mhagmajer:
Closes #129


YonatanKra:
Darn - I was in the middle of solving it ;)
I'll go over it during the evening.

czerwinskilukasz1:
@mhagmajer, let me check it within an hour.

src/askjsx/lib/Let.tsx
```diff
@@ -3,13 +3,18 @@ import * as askjsx from './jsx';
 import { AskCodeOrValue } from '../../askcode';
 askjsx;
 
-export function Set({
-  name,
-  value,
-}: {
+type Props = {
   name: string;
   type?: string;
```
YonatanKra, line 8: What is the "type" here?

src/askjsx/lib/Let.tsx
```diff
@@ -20,12 +25,13 @@ export function Set({
   );
 
   return (
-    <code let>
+    <code {...{ [code]: true }}>
```
YonatanKra, line 28: if code is just a string, isn't it enough to just print it here?

src/askvm/resources/core/let.ts
```diff
@@ -16,6 +16,21 @@ export const letRes = resource({
       throw new Error(`Key "resources" cannot be redeclared`);
     }
 
+    if (code.name === 'assign') {
+      for (
```
YonatanKra, line 20: It looks like you want to go into the prototype chain here. Are the values deep inside? In what case you'll have to search the values deeper than the first prototype?

YonatanKra:
Side note: reviewing this clearly shows we need to remove the auto-generated files from git. They are such a clatter... really hard to go over the PR.
Is that a thing? #170

src/askvm/resources/core/let.ts
```diff
@@ -1,4 +1,4 @@
-import { any, resource, run, runUntyped } from '../../lib';
+import { any, resource, run, runUntyped, Options } from '../../lib';
 
 export const letRes = resource({
```
YonatanKra, line 3: Can we give it a clearer name? Does `letRes` stand for `Let Resolve`?

src/askscript/__tests__/90-misc/scope-02.test.ts
```diff
@@ -0,0 +1,7 @@
+import program from './scope-02.ask';
```
YonatanKra, line 1: Why not have one test file for `let` and add all the possibilities there explicitly? 
It seems to me it is harder to maintain and much harder to understand the tests as they are now.
All in all, there are 2-3 scenarios per core functionality and it would be much easier to understand and follow if it were set in one file.
Something like:
```
import simpleAssign from './basics17-assign.ask';
import scopedAssign from './basics17-assign-scoped.ask';
import emptyAssign from './basics17-assign-empty.ask';

test('assign value for declared variable', async () => {
  await expect(await simpleAssign()).toEqual(2);
});

test('assign value inside scope for globally declared variable', async () => {
  await expect(await scopedAssign()).toEqual(2);
});

test('throw an error when assign to undeclared variable', async () => {
  await expect(emptyAssign()).rejects.toThrow();
});
```
@mhagmajer @czerwinskilukasz1 
What do you think?

src/askjsx/lib/Let.tsx
mhagmajer: in `let n: int = 5` it's the value of `int`

src/askjsx/lib/Let.tsx
mhagmajer: I don't knowa shorter way to add a new dynamic property in JSX tag

src/askvm/resources/core/let.ts
mhagmajer: If you are in a child scope (eg. inside the curly braces of an if block), your assignment should only affect the variable from your closest parent scope

src/askvm/resources/core/let.ts
mhagmajer: `letRes` actually stand for let resource. This pattern is used in other places as well. If we were to change it, let's make this a matter of another issue/PR.

src/askscript/__tests__/90-misc/scope-02.test.ts
mhagmajer: This is important and general subject. Let's take it into another issue. I also think that testing is very important and we should reconsider our approach here.

mhagmajer:
>Side note: reviewing this clearly shows we need to remove the auto-generated files from git. They are such a clatter... really hard to go over the PR.
Is that a thing? #170

Yes, @YonatanKra, this is being considered and we can discuss it further in its issue. Perhaps we don't need to keep parsing snapshots for all .ask files

src/askscript/__tests__/90-misc/scope-02.test.ts
czerwinskilukasz1: https://github.com/xFAANG/askql/issues/249

Issue #98, opened 2020-06-09
Implement tests for 'remote' block


czerwinskilukasz1:
Currently all tests for `remote` are marked as not implemented.
 ![image](https://user-images.githubusercontent.com/12752520/84160509-8c963480-aa6e-11ea-936f-392b52dfd632.png)

What is needed is running a server (or a mock server) during the tests which would respond on the code from the `remote` block.

czerwinskilukasz1:
Blocked on https://github.com/xFAANG/askql/issues/371

PR #204, opened 2020-06-21
fix(tests): remove false positives

### LINKED ISSUE #148: Fix false positive tests
Some test have invalid `expectedResult`. This issue is about auditing all tests and fixing those.

mhagmajer:
Close #148


czerwinskilukasz1:
@mhagmajer , how do we know these are all files where tests didn't work propertly?  I think there were some .ask test files, which were not constructed properly too.

mhagmajer:
@czerwinskilukasz1 these are all that I have found, we also have tests for remote that were disabled (tracked separately here: https://github.com/xFAANG/askql/issues/98). If you now about more we can add them here or in a new issue.

czerwinskilukasz1:
@mhagmajer , there was a number of tests you altered in 25 commits during your coding spree around 3am on May 25.
One of them is basics16c-const_without_value.ts.notImplemented from commit f228690006acd3f44887a72ecab9b9140b6ec320, which was supposed to fail when executing the .ask file, but you changed it to accepting it (see screenshot), which is incorrect. 

Please review all the commits created that night and resolve the issues.


image

czerwinskilukasz1:
@mhagmajer , what's the progress on this issue?

mhagmajer:
@czerwinskilukasz1 `const a` and `let b` are fine and they create a new value equal to `empty`

Issue #120, opened 2020-06-11
Expand Contributing guidelines by add Global .gitignore file configuration


SebastianWesolowski:
Add information in Contributing guidelines about Global .gitignore file configuration.

example set of ignore file:
https://gist.github.com/jacobtomlinson/aace65a6920e44348d501da5e13a5a17

~~My propose file extension is add specific macOS file:~~
~~add .DS_Store~~

czerwinskilukasz1:
Hi @SebastianWesolowski , good to see you here :) 
I fully support, would you like to submit a PR?

kgajowy:
@czerwinskilukasz1 Just wondering if this should happen. Repository .gitignore should relate only to given project artifacts (like dist), instead of what potential developers environment might add (like `.idea` depending on IDE and similar).
I know it is a common practice but it should be the case for every developer to have local settings of `.gitignore` to avoid bloating this file with hundreds of things not related to the repository (.DS_Store is OS specific not project, same as IDE-related things)

I would recommend not to add to `.gitignore` anything that does not belong to the project itself.

czerwinskilukasz1:
@kgajowy , thanks for your vigilance.  I totally forgot about the existence of .gitignore_global file!
Good point.

@SebastianWesolowski, kgajowy posted a sample minimal example above, which could work for you.
For a bigger set of ignores, take a look at: https://gist.github.com/jacobtomlinson/aace65a6920e44348d501da5e13a5a17
Do you see any reason for not using .gitignore_global for .DS_Store?

SebastianWesolowski:
@kgajowy yes you have right, it is a common practice. I usualy add exclude this file. 

global ignore is the best practice. What about people who don't have a configured globalignore file ? 

It was usually add exclusion better for a project to  than to suggest best practices. in my work, it worked.

I have no experience with a project where more than 4 developers have worked. Is that very important in project like this ?

I saw project like GraphQL use too similar practice like ignore file DS_Store, but not all repositroy from  https://github.com/graphql


https://github.com/graphql/graphql-spec/blob/master/.gitignore
https://github.com/graphql/graphiql/blob/master/.gitignore

mhagmajer:
I back up what @kgajowy said here. We can add gitignore global instructions in contributing.md

SebastianWesolowski:
I didn't look at the problem from this perspective. 

Writing instructions on how to configure global gitingore file is better solution.

I be happy to write that instruction. Shall I continue in this thread or create a new issue?

[...]

PR #219, opened 2020-06-23
Feat: Support both askcode and askscript in askExpressMiddleware

### LINKED ISSUE #191: demoServer.ts should accept AskScript instead of AskCode
The variable askScriptCode in https://github.com/xFAANG/askql/blame/master/src/askExpressMiddleware/askExpressMiddleware.ts#L20 suggests AskScript should be accepted but actually the server throws syntax errors for AskScript programs, while accepting AskCode.

YonatanKra:
Closes #191

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 node_modules
-/.idea
 /dist
 /false
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 node_modules
-
+/.idea
 /dist
 /false
 
--- a/src/askExpressMiddleware/askExpressMiddleware.ts
+++ b/src/askExpressMiddleware/askExpressMiddleware.ts
@@ -16,8 +16,8 @@ export function askExpressMiddleware(
     next: NextFunction
   ) {
     try {
-      const askScriptCode = request.body.code;
-      const queryResponse = await runUntyped(environment, parse(askScriptCode));
+      const askCode = request.body.code;
+      const queryResponse = await runUntyped(environment, parse(askCode));
       response.json(queryResponse);
     } catch (err) {
       if (config.passError) {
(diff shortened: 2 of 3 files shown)


src/askExpressMiddleware/askExpressMiddleware.ts
```diff
@@ -16,8 +16,8 @@ export function askExpressMiddleware(
     next: NextFunction
   ) {
     try {
-      const askScriptCode = request.body.code;
-      const queryResponse = await runUntyped(environment, parse(askScriptCode));
+      const askCode = request.body.code;
+      const queryResponse = await runUntyped(environment, parse(askCode));
```
mhagmajer, line 20: `parse` should come from the askscript package now, not askcode :)

mhagmajer: We'll support both AskCode and AskScript inputs (even without requiring user to provide an option) , but for now we need AskScript support for playground

mhagmajer:
Also please update the PR name according to conventional commits :)

src/askExpressMiddleware/askExpressMiddleware.ts
YonatanKra: @mhagmajer 
Please have a look now.
Also please look at the new test case. Is this a valid test for askscript?

.gitignore
```diff
@@ -1,5 +1,5 @@
 node_modules
-
+/.idea
```
mhagmajer, line 2: .globalignore

mhagmajer: See https://github.com/xFAANG/askql/issues/120 for reference

YonatanKra: Cool @mhagmajer ! Didn't know that :)
Solves many things in my many projects :)
So the code's working?

PR #220, opened 2020-06-23
feat: stats.steps and limits.steps in run options

### LINKED ISSUE #29: AskVM not timing out on infinite loops
According to the [FAQ](https://www.notion.so/FAQ-554ab8a775494057a0fdbedf7ad6f2d6), the AskVM should timeout on infinite loops.

However running the following in the AskVM generates a hang (infinite loop) that does not seem to resolve.

```
const f = fun():any {
  call(f)
}
call(f)
```

To replicate:
* `npm install askql`
* `./askql`
* `.editor`
* Enter the above code
* Ctrl+D

mhagmajer:
Close #29


src/askvm/index.ts
```diff
@@ -1,4 +1,32 @@
...
+  inputOptions: Options,
+  code: AskCodeOrValue,
+  args?: any[]
+): Promise<TypedValue<JSONable>> {
+  const options = extendOptions(inputOptions, {
+    stats: {
+      steps: 0,
+    },
+    limits: {
+      steps: 10 ** 4,
```
mhagmajer, line 20: When this was set to `10 ** 6` `while(true) {}` test run for almost 10 seconds

src/askvm/lib/run.ts
```diff
@@ -34,9 +39,9 @@ export async function run(
     return typed(code);
   }
 
-  ops += 1;
-  if (ops > 1000) {
-    // throw new Error('Over ops limit!');
+  options.stats!.steps += 1;
+  if (options.stats!.steps >= options.limits!.steps) {
+    throw new Error('Over steps limit!');
```
czerwinskilukasz1, line 44: @mhagmajer , could you list all operations that are considered a step?
It's important to make sure we don't interrupt valid computations too early.

mhagmajer: Every resource reference is a step. Also computing resources usually requires making more steps, for example:
- https://github.com/xFAANG/askql/blob/master/src/askvm/resources/core/if.ts
- https://github.com/xFAANG/askql/blob/master/src/askvm/resources/core/while.ts
- https://github.com/xFAANG/askql/blob/master/src/askvm/resources/core/query.ts

czerwinskilukasz1: OK, sounds good.
And how does fetching results in `query` count? 
E.g. if I wanted to get an aggregated query to sum people income in my corporation and run:
```
const people = query {
   person {
     income
   }
}
people:sum()
```
for a list of 10 001 people, would it fail?
How could I get the result I need anyway?

src/askvm/lib/run.ts
```diff
@@ -34,9 +39,9 @@ export async function run(
     return typed(code);
   }
 
-  ops += 1;
-  if (ops > 1000) {
-    // throw new Error('Over ops limit!');
+  options.stats!.steps += 1;
+  if (options.stats!.steps >= options.limits!.steps) {
```
czerwinskilukasz1, line 43: If think it should be > if we are saying we are *over* the limit.

src/askvm/lib/run.ts
mhagmajer: In this PR I only attempted to fix the issue with infinite loops. Configuring the limit will come next because it requires some changes to `extendOptions`

src/askvm/lib/run.ts
mhagmajer: This is at the beginning of the `run` operation, so if we are at the limit, then we cannot proceed.

src/askvm/lib/run.ts
czerwinskilukasz1: Configuring the limit is another topic. 
What I wanted to ask in the comment above is how `query` operations are counted?

czerwinskilukasz1:
> Let's add 2 more tests:
> One showing that running 10 ** 4 operations succeeds and
> one with very similar code showing that running 10 ** 4 + 1 operations fails

I think this one is still waiting for you.

mhagmajer:
@czerwinskilukasz1 let me understand better because these limits apply to all our passing tests. Would you like to have a loop that makes a lot of steps but still less than the limit?

czerwinskilukasz1:
> @czerwinskilukasz1 let me understand better because these limits apply to all our passing tests. Would you like to have a loop that makes a lot of steps but still less than the limit?

Yes, I would like to have:

- [ ] a test that almost reaches the ops limit (and finishes successfully), such that it's clearly visible in the code what limit it almost reach (e.g. an empty loop with 9 999 iterations, if the limit is 10 000)
- [ ] a test that would exceed the ops limit just a bit (and fails because of the limit), such that it's clearly visible in the code what limit it will reach (e.g. an empty loop with 10 000 iterations, if the limit is 10 000)

src/askvm/lib/run.ts
mhagmajer: I just checked that `query` uses `run` twice for each node - one for name and one for value. Maybe we can take a closer look at it and its performance on another issue once this is closed?

src/askvm/index.ts
```diff
@@ -1,4 +1,34 @@
...
+    },
+  });
+  const result = await libRun(options, code, args);
+  // console.log('stats after run', options.stats);
```
mhagmajer, line 24: leave it here to help with future debugging - this will eventually be replaced with a logging solution

src/askvm/lib/run.ts
czerwinskilukasz1: Sounds good.  Filing as https://github.com/xFAANG/askql/issues/303

src/askvm/index.ts
czerwinskilukasz1: ok, fine

src/askscript/__tests__/04-control-flow/program11-for_check_limit.test.ts
```diff
@@ -0,0 +1,9 @@
+import program from './program11-for_check_limit.ask';
+
+test('computes', async () => {
+  const maxArgUnderLimit = 587;
```
czerwinskilukasz1, line 4: Aaaah, so there is no obvious relationship between AskVM limit and number of iterations of an empty loop.

PR #223, opened 2020-06-25
Feat: Generate the package json in the dist folder on post build

### LINKED ISSUE #218: There is another package.json file in src/
This is our main package.json file:
https://github.com/xFAANG/askql/blob/master/package.json

And this is the package.json file I think we should remove:
https://github.com/xFAANG/askql/blob/master/src/package.json

YonatanKra:
- [x] Generate the package json in the dist folder on post build
- [x] Remove package.json from ts.config
- [x] Delete src/package.json

Closes #218


scripts/postBuild.ts
```diff
@@ -9,4 +30,19 @@ async function main() {
...
 }
 
+async function copyPackageJson() {
+  const releasePackageJson: any = {};
+  Object.keys(packageJson).forEach((key) => {
+    if (propertiesToPreserveFromPackageJSON.indexOf(key) > -1) {
+      // @ts-ignore
+      releasePackageJson[key] = packageJson[key];
+    }
+  });
```
mhagmajer, line 40: https://lodash.com/docs/4.17.15#pick

package.json
```diff
@@ -48,7 +48,9 @@
   "engines": {
     "node": ">=12.14.0"
   },
-  "dependencies": {},
+  "dependencies": {
+    "lodash": "^4.17.15"
+  },
```
mhagmajer, line 53: This needs to be dev dependency @YonatanKra

src/demoAskScriptServer.ts
```diff
@@ -14,8 +14,15 @@ import { parse as parseAskScript, AskScriptCode } from './askscript';
...
-import packageInfo from './package.json';
+let packageInfo = {
+  version: 'dummy',
+};
 
+try {
+  packageInfo = require('../package.json');
+} catch (e) {
+  packageInfo = require('./package.json');
+}
```
mhagmajer, line 25: Demo ask script server is currently only run as a development tool - we don't need to support it in a run package

PR #230, opened 2020-06-26
ci: enforce conventional-commits

### LINKED ISSUE #180: Enforce conventional commit in the message of squashed commit using GitHub actions

YonatanKra:
Closes #180 
I didn't use an action from the marketplace, because they all use `GITHUB_TOKEN` which isn't working with forks.
We can speed things up by creating a bundle with the node modules already in it to avoid the `npm i` phase in the action.
I thought it was an overkill since this runs much faster than the tests.


czerwinskilukasz1:
@YonatanKra , were you able to check that your solution works on Github?
What shows up to the user and when after they try to squash&merge with a message not conforming to conventional commits?

Are all of the types from [commitizen](https://github.com/commitizen/cz-cli) supported?
image

YonatanKra:
@czerwinskilukasz1 
This works according to these rules: https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional
As written on this page, they can be changed (for instance, don't enforce lowercase).

@mhagmajer 
What my solution is doing is not passing the PR if its title is not according to the standard. It's a linter for the PR title. Eventually, the PR title is used for the actual commit, so this should solve the issue, isn't it?

czerwinskilukasz1:
> @czerwinskilukasz1
> This works according to these rules: https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional
> As written on this page, they can be changed (for instance, don't enforce lowercase).

Sounds good to me, thanks.

PR #232, opened 2020-06-26
(STALE lambda PR)

### LINKED ISSUE #43: Handle providing a lambda function when chaining methods in AskScript
```
ask {
	const double = fun(s:string):string {
    s:concat(s)
	}

  text :call(fun(s:string): string {
    s:double // same as s:double()
  })
}
```
returns the value of `text` variable instead of `text` + `text`.

czerwinskilukasz1:
Closes #43


src/askscript/__tests__/03-lambda/program07b-chain_lambda.ask
```diff
@@ -1,5 +1,4 @@
 ask {
-  let useFor = fun(x, f) { f(x) }
```
mhagmajer, line 2: Okay so this definition is much simpler than [src/askvm/resources/core/useFor.ts](https://github.com/xFAANG/askql/pull/232/files#diff-ae653ac9c4d4704ef382af488f3b64f6). Why are we adding a new resource implemented in TypeScript instead of a new standard function implemented in AskScript?

czerwinskilukasz1: We discussed it [10 days ago](https://github.com/xFAANG/askql/issues/43#issuecomment-644754817) and I implemented it according to your comments.

mhagmajer:
This probably can be done on AskJSX level by defining a function - similar to how `for` is implemented

czerwinskilukasz1:
> This probably can be done on AskJSX level by defining a function - similar to how `for` is implemented

Doesn't it mean that useFor would need to be detected in AskScript parser and translated to 'useFor' structure there?

mhagmajer:
> Doesn't it mean that useFor would need to be detected in AskScript parser and translated to 'useFor' structure there?

No, you can define functions at AskJSX level - just add a special case to Call.jsx

czerwinskilukasz1:
Closing in favor of https://github.com/xFAANG/askql/pull/285

PR #234, opened 2020-06-26
feat(cli): add 'cli' and 'cli:dev' to npm run

czerwinskilukasz1:


package.json
```diff
@@ -19,7 +19,9 @@
     "prettier:check": "prettier --config prettier.config.js --check .",
     "prettier:scripts": "prettier --write ./scripts",
     "lint:js": "eslint . --ext .js,.tsx,.ts",
-    "lint:js:fix": "npm run lint:js --fix"
+    "lint:js:fix": "npm run lint:js --fix",
+    "cli": "node dist/cli.js",
+    "cli:dev": "npx pegjs src/askscript/parser/askscript.grammar.pegjs && ts-node --files src/cli.ts"
```
mhagmajer, line 24: 1. Where does `npx pegjs src/askscript/parser/askscript.grammar.pegjs` write its output?
2. Would ts-node src/cli.ts suffice here?

czerwinskilukasz1: 1. To `src/askscript/parser/askscript.grammar.js`, which is gitignored.
2. Nope, because src/askscript/index.ts says: `const parser: any = require('./parser/askscript.grammar');`. This file won't exist unless generated by peg.  We would need to embed peg here to make it work automagically, which is probably not the desired solution.

PR #245, opened 2020-06-26
feat(askvm): add logicalOr ('||')

### LINKED ISSUE #136: Add a 'logicalOr' resource (|| binary operator)
There are some operators in AskVM, e.g. 'lessThan' resource (< operator) (src/askvm/resources/math/lessThan.ts), but we are missing logical operators, like 'or' resource (|| operator).

Dear new contributor,

If you have any questions on how to start, please write a comment with a tag: @czerwinskilukasz1 or @mhagmajer, or join our Discord Community and let us know there. We will be very happy to help you start! :)

Cheers,
Łukasz (@czerwinskilukasz1)
AskQL Core Developer

czerwinskilukasz1:
Closes #136


mhagmajer:
I'm not sure if `||` should work for falsy values such as `0` or empty string. I think this is confusing. Why not just make it work for boolean?

czerwinskilukasz1:
> I'm not sure if `||` should work for falsy values such as `0` or empty string. I think this is confusing. Why not just make it work for boolean?

@mhagmajer 
Yeah, I was considering implementing it for boolean only, but then lines like `const port = process.env.PORT || 4000;`, which work perfectly fine in Javascript, won't work in AskScript.
What are your thoughts?

mhagmajer:
I think that is the use case for null coalescing operator actually `port ?? 4000`

czerwinskilukasz1:
> I think that is the use case for null coalescing operator actually `port ?? 4000`

I like it.   https://github.com/xFAANG/askql/issues/263

So agreed, `||` supports boolean only.

czerwinskilukasz1:
@mhagmajer , I am checking types explicitly. Let me know how you like it.

czerwinskilukasz1:
@mhagmajer  Great. I was totally unsure whether you would like it or hate it.

mhagmajer:
@czerwinskilukasz1 this is pretty much how I would do it today. These explicit type checks won't be needed when AskVM starts throwing on type mismatch

czerwinskilukasz1:
> @czerwinskilukasz1 this is pretty much how I would do it today. These explicit type checks won't be needed when AskVM starts throwing on type mismatch

@mhagmajer , sure.  When do you plan to implement type checks in AskVM?

mhagmajer:
1.3 is realistic I think

PR #259, opened 2020-06-27
feat(askscript): add syntax sugar for a[b]

### LINKED ISSUE #52: a.b / a[b] syntax sugar for a:at('b')

czerwinskilukasz1:
Closes #52


src/askscript/__tests__/08-objects-records/objects-15-key_brackets.ask
```diff
@@ -0,0 +1,6 @@
+ask {
+    let capitals:map(string) = { France: 'Paris', Germany: 'Berlin', Spain: 'Madrid', anywhere: 'any town' }
```
mhagmajer, line 2: Is `string` the type of key or value of the map? Perhaps we need two types for `map`. Also you actually an object (struct) to it which is implemented differently in JS than `Map`. Not sure if AskScript needs to make this differentiation...

czerwinskilukasz1: Currently map's keys are always strings, so the type describes the values.

mhagmajer: @czerwinskilukasz1 this is true for AskScript, not AskCode

czerwinskilukasz1: https://github.com/xFAANG/askql/issues/261

PR #271, opened 2020-06-27
feat(tests): add expect and toBe resources

### LINKED ISSUE #154: Add `test` and `expect` resources for fast and concise testing 🔧
The following program should fail when run in CLI, however it succeeds. 
Moreover, the log() function doesn't seem to be called. I assume the whole function to test is not run at all.

This is a fast and concise way to define tests, it would be great if this worked soon.

```
🦄 .editor
// Entering editor mode (^D to finish, ^C to cancel)
  test('hello', fun () {
    log('Hello!')

    expect(2 + 2):toBe(4)
    expect(2 + 2):toBe(20)
  })

int ask(call(get('test'),'hello',fun(call(get('log'),'Hello!'),call(get('toBe'),call(get('expect'),call(get('+'),2,2)),4),call(get('toBe'),call(get('expect'),call(get('+'),2,2)),20))))
5
🦄 
```

### LINKED ISSUE #289: Treat AskScript tests (.test.ask) like TypeScript tests (.test.ts)
Following mhagmajer's [comment](https://github.com/xFAANG/askql/pull/271#discussion_r446969779), AskScript tests (.test.ask) should be treated more like TypeScript tests (.test.ts) - we shouldn't rely on .test.result.ts and test.args.ts files for them

czerwinskilukasz1:
Closes #154 
Closes #289 

What I managed to add is `expect` and `toBe` resources, however `test()` resource is not defined, despite that it's _declared_ in `index.d.ts` file.  Also, `test` (and other test-related resources) are missing in `jestEnvironment.global` dictionary. 
It's possible that extra steps are needed, for example adding a piece of code like:
```
  const jasmineFactory = runtime.requireInternalModule<
    typeof import('./jasmine/jasmineLight')
  >(JASMINE);
  const jasmine = jasmineFactory.create({
    process,
    testPath,
    testTimeout: globalConfig.testTimeout,
  });

  const env = jasmine.getEnv();
  const jasmineInterface = jasmineFactory._interface(jasmine, env);
  Object.assign(environment.global, jasmineInterface);
```
(source: https://github.com/facebook/jest/blob/master/packages/jest-jasmine2/src/index.ts#L44)

This code snippet is ugly, so I didn't dare to add it to our code.  I suggest we just content ourselves with `expect` and `toBe`.

Here is how an error looks - as if it was a pure jest:
Screen Shot 2020-06-27 at 18 23 18


package.json
```diff
@@ -88,6 +89,7 @@
     "eslint": "7.3.0",
     "eslint-config-prettier": "6.11.0",
     "eslint-plugin-prettier": "3.1.4",
+    "expect": "^26.1.0",
```
mhagmajer, line 92: Everything we should need is shipped with Jest already. This isn't ideal to add another dependency here, but maybe this is the easiest way to start

src/askscript/__tests__/00-documentation-examples/documentation01-complete_example.test.ask
```diff
@@ -1,9 +1,7 @@
 ask {
-  test('hello', fun () {
```
mhagmajer, line 2: Why not keep the `test` resource - It's a standard of Jest to group expectations into tests

package.json
mhagmajer, line 92: Everything we should need is shipped with Jest already. This isn't ideal to add another dependency here, but maybe this is the easiest way to start. I've also tried to use `expect` and `test` from jest internal environment and it was not easy

src/askscript/__tests__/00-documentation-examples/documentation01-complete_example.test.test.result.ts
```diff
@@ -0,0 +1 @@
+export = null;
```
mhagmajer, line 1: AskScript tests (.test.ask) should be treated more like TypeScript tests (.test.ts) - we shouldn't rely on .test.result.ts and test.args.ts files for them

src/test.jest.testRunner.ts
```diff
@@ -164,8 +166,26 @@ async function askRunner(
...
+      },
+    }),
+
+    toBe: resource({
+      type: any,
+      async resolver(matcher: e.Matchers<any>, actual: any): Promise<void> {
+        return matcher.toBe(actual);
+      },
+    }),
+  };
```
mhagmajer, line 185: Can we add more? The are all going to be implemented in a similar way - so this shouldn't require to much work without code repetition

src/test.jest.testRunner.ts
```diff
@@ -164,8 +166,26 @@ async function askRunner(
     title: 'compiles',
   });
 
+  // Adding Jest functions to the environment, so that we can run Jest expect():toBe() in .ask files.
+  // See a sample in src/askscript/__tests__/00-documentation-examples/documentation01-complete_example.test.ask
+  const testResources = {
+    expect: resource({
+      type: any,
+      async resolver(actual: any): Promise<e.Matchers<any>> {
+        return e(actual);
```
mhagmajer, line 175: Please add a comment explaining that ideally we would reuse `test` and `expect` from the Jest environment instead of a separate package

src/askscript/__tests__/00-documentation-examples/documentation01-complete_example.test.ask
mhagmajer: You could optionally use `createEmptyAssertionResult`, `passed`, `failed` function to mark individual tests. Let's keep it easy as this is a prototype integration between AskScript <-> Jest, however, let's support all the basic blocks such us `test`, `expect` and `toBe` so that we can start writing AskScript tests already and then figure out better Jest integration

mhagmajer:
Ideally your screenshot would show the problematic AskScript code. That's probably something that will require more iterations on this integration. For this PR let's keep it simple

src/askscript/__tests__/00-documentation-examples/documentation01-complete_example.test.ask
czerwinskilukasz1: > Why not keep the `test` resource - It's a standard of Jest to group expectations into tests

@mhagmajer , I tried to add the Jest's test resource, but failed. See https://github.com/xFAANG/askql/pull/271#issue-440937327 for details.

czerwinskilukasz1: > You could optionally use `createEmptyAssertionResult`, `passed`, `failed` function to mark individual tests. Let's keep it easy as this is a prototype integration between AskScript <-> Jest, however, let's support all the basic blocks such us `test`, `expect` and `toBe`/`toEqual` etc so that we can start writing AskScript tests already and then figure out better Jest integration

Current PR is enough for writing AskScript tests, see screenshot at https://github.com/xFAANG/askql/pull/271#issue-440937327

package.json
czerwinskilukasz1: I think they don't expose the classes in their .d.ts files, that's why the standard `jest` package was not enough.

src/askscript/__tests__/00-documentation-examples/documentation01-complete_example.test.test.result.ts
czerwinskilukasz1: Let's keep the PRs small.  Filing your note as https://github.com/xFAANG/askql/issues/289

src/test.jest.testRunner.ts
czerwinskilukasz1: Of course we could.  Let's discuss which resources to add: https://github.com/xFAANG/askql/issues/290

src/test.jest.testRunner.ts
czerwinskilukasz1: Done.

czerwinskilukasz1:
> Ideally your screenshot would show the problematic AskScript code. That's probably something that will require more iterations on this integration. For this PR let's keep it simple

Good idea for future enhancements. Filing it as https://github.com/xFAANG/askql/issues/291

src/askscript/__tests__/00-documentation-examples/documentation01-complete_example.test.ask
mhagmajer: @czerwinskilukasz1 can you add very simple implementation of `test`? I'd like AskScript tests to group tests same way like TypeScript does. It can have very simple implementation like:

```
let test = fun(name, f) { f() }
```

package.json
mhagmajer: Not sure, definitely more research is needed on more native Jest integration

src/askscript/__tests__/00-documentation-examples/documentation01-complete_example.test.test.result.ts
mhagmajer: @czerwinskilukasz1 this I think it makes better sense to fix here, this PR is an enhancement, not a bug fix - so let's make sure that we are stepping in the right direction from the first step

src/test.jest.testRunner.ts
mhagmajer: Sure, we can do it as a follow up

czerwinskilukasz1:
> @czerwinskilukasz1 thanks for creating issues for follow ups. I do think actually that these two points should be fixed here:
> 
> * [ ]  allow .test.ask to run independently (no .test.result.ts etc)

By the way, if we allow <filename>.test.ask to run independently, shouldn't <filename>.test.result.ts and <filename>.test.args.ts be renamed to <filename>.result.ts and <filename>.args.ts, because they are used for <filename>.ask and not for <filename>.test.ask?

src/askscript/__tests__/00-documentation-examples/documentation01-complete_example.test.ask
czerwinskilukasz1: @mhagmajer , I guess we don't have a good way to define resources in AskScript, do we?

czerwinskilukasz1: I think I don't understand the way askvm calls functions well enough. 
I implemented `test` as:
```
test: resource({
      type: any,
      async compute(options, code, args) {
        assert(
          typeof args !== 'undefined' && args.length == 2,
          'test() expects exactly two arguments'
        );
        const fun = args[1] as TypedValue<AskCode>;
        return call.compute(options, fun.value);
      },
    }),
```

and I'm getting 'Cannot call null'.
Any tips?

mhagmajer:
> By the way, if we allow .test.ask to run independently, shouldn't .test.result.ts and .test.args.ts be renamed to .result.ts and .args.ts, because they are used for .ask and not for .test.ask?

Maybe, let's figure this out on a separate issue

czerwinskilukasz1:
> > By the way, if we allow .test.ask to run independently, shouldn't .test.result.ts and .test.args.ts be renamed to .result.ts and .args.ts, because they are used for .ask and not for .test.ask?
> 
> Maybe, let's figure this out on a separate issue

Filing as https://github.com/xFAANG/askql/issues/293

PR #285, opened 2020-06-29
feat(askvm): add useFor resource (lambda function) 

### LINKED ISSUE #43: Handle providing a lambda function when chaining methods in AskScript

czerwinskilukasz1:
Closes #43


src/askscript/__tests__/03-lambda/program07c-custom_lambda.ask
```diff
@@ -0,0 +1,9 @@
+ask {
+	let customUseFor = fun(x, f) { f(x) }
```
mhagmajer, line 2: :+1:

src/askjsx/lib/Call.tsx
```diff
@@ -15,6 +15,21 @@ export function Call({
 }) {
   assert(isString(name), 'name');
   // assert(isStringArray(args), 'args');
+  if (name === 'useFor') {
```
mhagmajer, line 18: Please add a comment explaining that this is a temporary way of linking some AskScript code. We'll have to make typing work before introducing a better way

src/askjsx/lib/Call.tsx
```diff
@@ -15,6 +15,24 @@ export function Call({
...
+
+  // This is a temporary way of linking some AskScript code.
+  // We'll have to make typing work before introducing a better way
+  if (name === 'useFor') {
+    assert('params' in args, 'No params found in arguments');
+    const params = (args as any).params;
+    assert(Array.isArray(params), 'params should be an array');
+    assert(params.length == 2, `expecting exactly 2 params for ${name}`);
+
+    if ((args as any).params)
```
mhagmajer, line 27: this check is not needed

src/askjsx/lib/Call.tsx
```diff
@@ -15,6 +15,24 @@ export function Call({
...
+      return (
+        <code call>
+          <Ref name={'call'} />
+          {[(args as any).params[1], (args as any).params[0]]}
```
mhagmajer, line 31: why not just use `params` here?

czerwinskilukasz1: Right

PR #297, opened 2020-06-30
feat(askvm): add forOf resource + tests

### LINKED ISSUE #132: for...of doesn't seem to execute the body
Executing the following program:
```
ask {
    let mySum:float = 0
    for(let price of itemPrices) {
        mySum = mySum:plus(price)
    }
    mySum
}
```
should lead to an error as itemPrices does not exist and is not defined as resource or value either.
Instead, this program return 0.

czerwinskilukasz1:
Closes #132


czerwinskilukasz1:
🚀

PR #300, opened 2020-06-30
feat(askvm): implement for..in + tests

### LINKED ISSUE #131: for...in doesn't seem to execute the body
Executing the following program:
```
ask {
    let mySum:float = 0
    for(let index in itemPrices) {
        const price = itemPrices:at(index)
        mySum = mySum:plus(price)
    }
    mySum
}
```

should lead to an error as `itemPrices` does not exist and is not defined as resource or value either.
Instead, this program returns 0.

czerwinskilukasz1:
Closes #131


src/askvm/resources/core/forIn.ts
```diff
@@ -1,11 +1,52 @@
...
+    }
+    if (!isAskCode(key)) {
+      throw new Error('forIn expects key to be a statement');
+    }
+    if (!['let', 'const'].includes(key.name)) {
+      throw new Error('forIn expects key to be a "let" or "const" statement');
+    }
+    if (!Array.isArray(key.params)) {
+      throw new Error('forIn expects key to have params!');
+    }
```
mhagmajer, line 23: Nice error handling 👍

src/askvm/resources/core/forIn.ts
```diff
@@ -1,11 +1,52 @@
...
+      );
+    }
+
+    const values: any[] =
+      typeof valueOf === 'object'
+        ? Object.keys(valueOf)
+        : Array.from({ length: valueOf.length }, (value, key) => key);
+
+    for (const value of values) {
+      key.params[1] = value;
```
mhagmajer, line 43: it's a little hacky to modify the param like here, but it's okay for the first implementation fo this resource.

czerwinskilukasz1: Thanks. By the way, I think I did similar for for of.

src/askvm/resources/core/forIn.ts
czerwinskilukasz1: Thanks :) It would be nice to have good error handling in the whole Ask.

czerwinskilukasz1:
> GIven that we already have for of, I wonder if
> 
> ```
> for (let key in object) { ... }
> ```
> 
> Can be simply translated as
> 
> ```
> for (let key of object:keys) {
> ```
> 
> Because if so, maybe we can reuse the forOf implementation more?

Technically, we could, _but_ we would need a couple of extra lines of code for handling a string.
Given that this is the first version, let's keep it this way for now. If we end up _not_  implementing it in AskScript anytime soon, we could rework it then.

PR #311, opened 2020-07-01
feat(askscript): add unary operators

### LINKED ISSUE #298: -n / -(n) doesn't parse
```
🦄 -test
Uncaught Error [SyntaxError]: Expected [0-9] but "t" found.
Location: {
  "start": {
    "offset": 7,
    "line": 2,
    "column": 2
  },
  "end": {
    "offset": 8,
    "line": 2,
    "column": 3
  }
}
```

Expected: negative value of `test`

If we reworked prefix unary operators, then it would help in other areas such as `+4` which also doesn't parse. If unary minus was an operator rather than part of number's grammar rule we could solve the plus problem very easily.

Tip from JS: `2++4` doesn't work, but `2+(+3)` does

czerwinskilukasz1:
Closes #298


czerwinskilukasz1:
@mhagmajer , note that I needed to add extra steps to handle unary operator priority, so that:
`-3 < -2` is equivalent to: `(-3) < (-2)` instead of `-(3 < -2)`
`-3:greaterThan(-2)` is equivalent to: `(-3):greaterThan(-2)` instead of `-(3:greaterThan(-2))`

src/askscript/__tests__/01-basics/program03d-int.askcode
```diff
@@ -1 +1 @@
-ask(123,-456)
\ No newline at end of file
+ask(123,456)
```
mhagmajer, line 1: What happened to the minus? 🤔

src/askvm/resources/math/minus.ts
```diff
@@ -3,6 +3,12 @@ import { any, resource } from '../../lib';
 export const minus = resource({
   type: any,
   async resolver(a: number, b: number): Promise<number> {
-    return a - b;
+    if (typeof b === 'undefined') {
```
mhagmajer, line 6: Let's update the TypeScript argument type for `b` too

czerwinskilukasz1: Good catch!   
I wonder why tsc doesn't generate any warning in such cases.

src/askscript/__tests__/01-basics/program03d-int.askcode
czerwinskilukasz1: I will try to figure it out, but if it takes too long, how about you help me?

mhagmajer: Sure, let's figure this out together 🙌

czerwinskilukasz1: What I discovered so far is it has to do with Prettier:
Screen Shot 2020-07-02 at 12 46 51

Let me do more research.

mhagmajer: @czerwinskilukasz1 yes, that makes sense. Prettier needs to be aware of new constructs we're adding to be able to produce correctly formatted code

mhagmajer: And .askcode is currently reflection of the formatted code to make sure that prettier layer also gets tested

czerwinskilukasz1: The issue is caused by the assumption that all operators are binary.  So `join` expects 2 arguments and wants to add an operator between them here:
image
However, for unary operators there is obviously only one operand.

czerwinskilukasz1: Fixed

Issue #296, opened 2020-06-30
unequality (operator != or <>) is missing


czerwinskilukasz1:
```
ask {
    if (2<>2) { return true}
}
```
results in:
```
Error: Unknown identifier '<>'!
```

The task is to implement unequality. 

Some discussion is needed to decide whether we add `!=` or rather `<>`. Depends on discussion in https://github.com/xFAANG/askql/issues/295 .

mhagmajer:
I guess `!=` is known better :)

czerwinskilukasz1:
Blocked on #295

czerwinskilukasz1:
I guess != should call not(equals()), right, @mhagmajer ?

mhagmajer:
@czerwinskilukasz1 yes, sure

czerwinskilukasz1:
@mhagmajer , we have a number of operators defined in src/askvm/resources/math/index.ts and I think it would be the most logical to place `!=` there. 
However, I can't figure out how to define `!=` as `not(equals())` there. Would I need to implement a separate resource? How could I refer to `not` and `equals` resources there?

Another option would be to define `!=` somewhere in askjsx, but I guess I would need to write an ugly if in `src/askjsx/lib/Call.tsx` which would support just `!=` and no other operators.  I'm not a fan of this solution.

What's your advice?

mhagmajer:
@czerwinskilukasz1 let's just create `notEquals` resource for now and then simplify it later on

czerwinskilukasz1:
Blocked on #295

PR #342, opened 2020-07-06
feat(askscript): add exclamation mark (!) to operators (#295)

### LINKED ISSUE #295: Add '!' to valid characters for operators
When I run:
```
ask {
    if (2!=2) { return true}
}
```

I got:
```
)
ask {
    if (2!=2) { return true}


SyntaxError: Expected " ", ")", ",", ".", ":", "\n", "\r", "\t", [/], [0-9], or [\-<>+*\^%=&|] but "!" found.
Location: {
  "start": {
    "offset": 15,
    "line": 2,
    "column": 10
  },
  "end": {
    "offset": 16,
    "line": 2,
    "column": 11
  }
}
```

Shouldn't we add '!' to valid characters for operators?

czerwinskilukasz1:
Closes #295

### DIFF
--- a/src/askscript/__tests__/13-real-operators/real-operators-01e-exclamation.ask
+++ b/src/askscript/__tests__/13-real-operators/real-operators-01e-exclamation.ask
@@ -1,0 +1,3 @@
+ask { // Implemented in https://github.com/xFAANG/askql/issues/295
+    if (2!=2) { return true}
+}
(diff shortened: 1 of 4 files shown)


src/askscript/__tests__/13-real-operators/real-operators-01e-exclamation.ask
```diff
@@ -0,0 +1,3 @@
+ask { // Implemented in https://github.com/xFAANG/askql/issues/295
```
mhagmajer, line 1: Please add more test cases, eg.

`a != 0 && a != 1`

And add expected test result

czerwinskilukasz1: This would be a good test for https://github.com/xFAANG/askql/issues/296 (Add != operator) but the current ticket was only about adding '!' to the operators in the grammar, so the only good test is to check that .ask file parses successfully (and it does).

Issue #323, opened 2020-07-03
Deploy Playground frontend on Netlify


czerwinskilukasz1:
This will also solve frontend part of #283 (HTTPS).

czerwinskilukasz1:
The code is in PR #327

czerwinskilukasz1:
Blocked on #345

czerwinskilukasz1:
This is done now.

Issue #235, opened 2020-06-26
Add Google Analytics to Playground frontend


czerwinskilukasz1:
This would help us see analytics, how many users use Playground and how long they spend on cli.askql.org

PR #348, opened 2020-07-06
fix(playground): add GA to the site for Netlify too

czerwinskilukasz1:
Related to #235 and #323


czerwinskilukasz1:
@mhagmajer , could you review please?

Issue #370, opened 2020-07-08
[askvm] 'fetch' resource should not download a hardcoded URL


czerwinskilukasz1:
Currently, [fetch resource](https://github.com/xFAANG/askql/blob/master/src/askvm/resources/node/fetch.ts) always downloads https://jsonplaceholder.typicode.com/posts/1 instead of an URL from the argument.

RichardM99:
@czerwinskilukasz1 I can take if that's ok? 🙃

czerwinskilukasz1:
Hello Richard, you got it! 😃

czerwinskilukasz1:
@RichardM99 , out of curiosity, how did you find AskQL repository? We are working on making it a bit more popular and I was wondering which way worked in your case.

RichardM99:
@czerwinskilukasz1 

Thank you 🙃 

It came up on my twitter feed because I follow https://twitter.com/first_tmrs_only

czerwinskilukasz1:
Great, thanks for feedback :) 
Also, feel free to give feedback to *us*. There are probably tons of things we could improve.  You can write here or via a message on our Discord: https://discord.gg/pYdzypH

Issue #269, opened 2020-06-27
toInt doesn't convert numbers to integer


czerwinskilukasz1:
```
🦄 2.3:toInt
float ask(call(get('toInt'),2.3))
2.3
🦄 
```

Dear Contributor,

This task has evolved a bit. Please read the comments to find out the latest description.

czerwinskilukasz1:
@mhagmajer , was your intention to create a real `toInt` or rather `toNumber`?

mhagmajer:
@czerwinskilukasz1 this should round floats to be integers, I confirm that this is a bug in implementation of `toInt`

czerwinskilukasz1:
@mhagmajer 
Actually, I wasn't sure because we already have `floor` and `ceil`:

```
🦄 floor(2.3)
int ask(call(get('floor'),2.3))
2
🦄 ceil(2.3)
int ask(call(get('ceil'),2.3))
3
🦄 
```

So maybe we could just remove `toInt` entirely and introduce `toNumber` instead, which would convert any _reasonable_ type to int | float?
```
toNumber('2.3')  = 2.3 (float)
toNumber(2.3) = 2.3 (float)
toNumber('2') = 2 (int)
toNumber(2) = 2 (int)
toNumber('a') = NaN
toNumber([]) = NaN

mhagmajer:
I'd like to live it up to user to decide to if they want int or float because these types will be dealt with differently by runtime

mhagmajer:
`floor` and `ceil` should only work on numbers

czerwinskilukasz1:
OK, tickets added 😄

czerwinskilukasz1:
So back to `toInt`, should:
 - `toInt(2.6)` return `2` or `3`?
 - `toInt('2.6')` return `2` or `3`?
 - `toInt(true)` return `1`?
 - `toInt(false)` return `0`?
 - `toInt(null)` return `0`?
 - `toInt([])` return `0`?
 - `toInt([1])` throw?
 - `toInt([1,2])` throw?
 - `toInt({a:1})` throw?
 - `toInt({a:1, b: 2})` throw?

czerwinskilukasz1:
I'd think it's a good default, since programmers are used to it and at the same time I don't see big no's against.

mhagmajer:
Let's throw on complex types such is array or struct. Not all strings should be supported to

czerwinskilukasz1:
[...]

Issue #181, opened 2020-06-17
Add more examples to the AskScript Playground


czerwinskilukasz1:
The [Askscript Playground](http://cli.askql.org:3000/) has a couple of examples not but it will be great to add there 15-20 more, half of them showing different features of the language and half of them being complete programs doing "real things".

czerwinskilukasz1:
My idea is to write in AskScript solutions for some coding tasks for Javascript/Typescript from https://www.codewars.com/ or a similar website.

PR #381, opened 2020-07-09
fix(askvm): add url param to fetch

RichardM99:
- Resolves https://github.com/xFAANG/askql/issues/370

### DIFF
--- a/src/askscript/__tests__/90-misc/fetch.test.ts
+++ b/src/askscript/__tests__/90-misc/fetch.test.ts
@@ -1,41 +1,43 @@
-jest.mock('node-fetch');
-import fetch from 'node-fetch';
-import { resources } from '../../../askvm';
-import { e2e } from '../../../utils/tools';
+import program from './fetch.ask';
 
 describe('fetch.ask', () => {
-  let result: any;
-
-  beforeAll(async () => {
-    fetch.mockReturnValue(mockedResponse());
-    const environment = { resources };
-    result = await e2e(
-      "ask { fetch('https://jsonplaceholder.typicode.com/posts/1') }",
-      environment
-    );
+  beforeEach(() => {
+    fetch.resetMocks();
   });
 
-  it('should return expected data', () => {
-    expect(result).toEqual({ id: 1 });
+  it('should return expected data', async () => {
+    fetch.mockResponse(Promise.resolve(response()));
+
+    const result = await program();
+
+    expect(result.id).toEqual(2);
   });
 
-  it('should call node fetch the correct number of times', () => {
-    expect(fetch).toHaveBeenCalledTimes(1);
+  it('should be called the correct number of times', async () => {
+    fetch.mockResponse(Promise.resolve(response()));
+
+    await program();
+
+    expect(fetch.mock.calls.length).toEqual(1);
   });
 
-  it('should call node fetch with the correct url', () => {
-    expect(fetch).toHaveBeenCalledWith(
+  it('should call the correct url', async () => {
+    fetch.mockResponse(Promise.resolve(response()));
+
+    await program();
+
+    expect(fetch.mock.calls[0][0]).toEqual(
       'https://jsonplaceholder.typicode.com/posts/1'
     );
   });
 });
 
-function mockedResponse() {
-  return Promise.resolve({
+function response() {
+  return {
     json: () => {
       return {
-        id: 1,
+        id: 2,
       };
     },
-  });
+  };
 }
--- a/src/askscript/__tests__/90-misc/fetch.test.ts
+++ b/src/askscript/__tests__/90-misc/fetch.test.ts
@@ -1,43 +1,41 @@
-import program from './fetch.ask';
+jest.mock('node-fetch');
+import fetch from 'node-fetch';
+import { resources } from '../../../askvm';
+import { e2e } from '../../../utils/tools';
 
 describe('fetch.ask', () => {
-  beforeEach(() => {
-    fetch.resetMocks();
+  let result: any;
+
+  beforeAll(async () => {
+    fetch.mockReturnValue(mockedResponse());
+    const environment = { resources };
+    result = await e2e(
+      "ask { fetch('https://jsonplaceholder.typicode.com/posts/1') }",
+      environment
+    );
   });
 
-  it('should return expected data', async () => {
-    fetch.mockResponse(Promise.resolve(response()));
-
-    const result = await program();
-
-    expect(result.id).toEqual(2);
+  it('should return expected data', () => {
+    expect(result).toEqual({ id: 1 });
   });
 
-  it('should be called the correct number of times', async () => {
-    fetch.mockResponse(Promise.resolve(response()));
-
-    await program();
-
-    expect(fetch.mock.calls.length).toEqual(1);
+  it('should call node fetch the correct number of times', () => {
+    expect(fetch).toHaveBeenCalledTimes(1);
   });
 
-  it('should call the correct url', async () => {
-    fetch.mockResponse(Promise.resolve(response()));
-
-    await program();
-
-    expect(fetch.mock.calls[0][0]).toEqual(
+  it('should call node fetch with the correct url', () => {
+    expect(fetch).toHaveBeenCalledWith(
       'https://jsonplaceholder.typicode.com/posts/1'
     );
   });
 });
 
-function response() {
-  return {
+function mockedResponse() {
+  return Promise.resolve({
     json: () => {
       return {
-        id: 2,
+        id: 1,
       };
     },
-  };
+  });
 }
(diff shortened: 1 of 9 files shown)


czerwinskilukasz1:
Hello, Richard! Nice to see you here and thank you for your PR!
I am wondering if you had a chance to check that the code works as intended after your change.
The best way would be to create an .ask file with a test that calls for example `fetch('https://jsonplaceholder.typicode.com/posts/1')` (jsonplaceholder.typicode.com is a website with dummy sample data) and checks whether a value for 'id' key in the received data equals to `1`.

@RichardM99

RichardM99:
@czerwinskilukasz1 

Thanks for the feedback - will do!

RichardM99:
@czerwinskilukasz1 

I'm afraid I don't understand the test set up well enough to add a passing test.  I added a test (based it on the `scope-02` test) but it throws type error "fetchFun is not a function".

Also not sure whether the test should mock the http call or actually make it.

src/askscript/__tests__/90-misc/fetch.test.ts
```diff
@@ -0,0 +1,6 @@
+import program from './fetch.ask';
+
+test('fetch', async () => {
+  const json = await program();
+  expect(json.id).toBe(1);
+});
```
mhagmajer, line 6: @RichardM99 can we mock the endpoint for tests? The tests should run in isolation. You can find more about mocking in Jest here https://jestjs.io/docs/en/mock-functions

mhagmajer: It's also useful to hit the real endpoint, however, that would be an integration test for Node which we don't yet have - we currently focus on unit test coverage

czerwinskilukasz1:
> @czerwinskilukasz1
> 
> I'm afraid I don't understand the test set up well enough to add a passing test. I added a test (based it on the `scope-02` test) but it throws type error "fetchFun is not a function".

@RichardM99 , congrats, it seems you have found a bug after fixing a bug :) 
If there had been no second bug, `fetchFun` would have just worked.
How about you try to figure out why it didn't work? This Stack Overflow thread might be helpful: https://stackoverflow.com/questions/48433783/referenceerror-fetch-is-not-defined

As for mock tests which @mhagmajer suggested, let's focus on making the code compile first.

src/askscript/__tests__/90-misc/fetch.test.ts
mhagmajer: I'm going to let @czerwinskilukasz1 to continue guide you through your first contributions :)

RichardM99:
@czerwinskilukasz1 

Tried both `global.fetch = require('node-fetch')` as well as just requiring it directly in `node/fetch.ts` but same error. Also not clear to me why the error is always `fetchFun` is not a function - even when I change the name of the function and `fetchFun` is not existing in the code base.

czerwinskilukasz1:
> @czerwinskilukasz1
> 
> Tried both `global.fetch = require('node-fetch')` as well as just requiring it directly in `node/fetch.ts` but same error. Also not clear to me why the error is always `fetchFun` is not a function - even when I change the name of the function and `fetchFun` is not existing in the code base.

It's an interesting puzzle, let me take a look.

czerwinskilukasz1:
> > @czerwinskilukasz1
> > Tried both `global.fetch = require('node-fetch')` as well as just requiring it directly in `node/fetch.ts` but same error. Also not clear to me why the error is always `fetchFun` is not a function - even when I change the name of the function and `fetchFun` is not existing in the code base.
> 
> It's an interesting puzzle, let me take a look.

OK, so I made the test work:
image

What I did is:
1. I added the following line:
image

2. Recompiled (`npm run build`) and only after recompiling did I run `npm test`.   I am guessing you just run `npm test` straightaway, which is what I sometimes do too and discover things are still failing.

OK, so now the last step.
Having:
```
global.fetch = require('node-fetch');
const fetchFun = (global as any).fetch;
```
doesn't look nice. First, we tend to use `import` construct instead of `require`. Second, I would just skip the assignment to global, so that a file deep inside the code doesn't pollute the global namespace, and would just use a local variable.

How about that, @RichardM99 ?

RichardM99:
@czerwinskilukasz1 

That's great - thank you 🙃 

I will make the required changes over the weekend.

RichardM99:
@czerwinskilukasz1 

Hello!

So as per your suggestions was able to make the tests pass.

However - now running into the issue of being unable to mock node-fetch.

Tried 

https://jestjs.io/docs/en/bypassing-module-mocks

And as you will see in my latest changes I also tried https://www.npmjs.com/package/jest-fetch-mock

Unfortunately the actual API call is always made instead of the mock data being resolved.

czerwinskilukasz1:
> @czerwinskilukasz1
> 
> Hello!
> 
> So as per your suggestions was able to make the tests pass.

Good job!

> 
> However - now running into the issue of being unable to mock node-fetch.
> 
> Tried
> 
> https://jestjs.io/docs/en/bypassing-module-mocks
> 
> And as you will see in my latest changes I also tried https://www.npmjs.com/package/jest-fetch-mock
> 
> Unfortunately the actual API call is always made instead of the mock data being resolved.
> 
> EDIT: Perhaps instead of importing `node-fetch` locally in `fetch.ts` - it should be assigned globally?

Yeah, mocking can be tricky in some cases.
I like the tests you wrote though.
I couldn't make jest-fetch-mock work, it might need some extra tweaks or we are both missing something.
Anyways, I made your *first test* pass by following the example you linked: https://jestjs.io/docs/en/bypassing-module-mocks **and** in fetch.test.ts:

1. replacing
```
const result = await program();
```

with:
```
    const environment = { resources };
    const result = await e2e(
      "ask { fetch('https://jsonplaceholder.typicode.com/posts/1') }",
      environment
    );
```
and 

2. adding the following headers:
```
jest.mock('node-fetch');
import fetch from 'node-fetch';
import { resources } from '../../../askvm';
import { e2e } from '../../../utils/tools';
```

Ah, and because I used jest.mock() within fetch.test.ts, I removed setupJest.js

Why your solution didn't work?  I *suspect* importing AskQL using `import program from './fetch.ask';` uses some Jest import mechanisms and the file was read and compiled in a separate context, where fetch was not mocked.  @mhagmajer wrote the mechanism and he might be able to add more context. 
Anyways, how about you try to make my solution work for 2nd and 3rd tests too?

RichardM99:
@czerwinskilukasz1 

Hey thanks for the comments! For sure would not have figured out to use the `e2e` function without support 🙃 

Applied your comments! Let me know what you think (maybe it's a bit weird to now have an `e2e` function in the unit test?)

czerwinskilukasz1:
> @czerwinskilukasz1
> 
> Hey thanks for the comments! For sure would not have figured out to use the `e2e` function without support 🙃
> 

Of course, you're welcome 🙂 

> Applied your comments! Let me know what you think (maybe it's a bit weird to now have an `e2e` function in the unit test?)

As for `e2e` function, the name didn't mean to mean end-to-end testing but rather compiling and running end-to-end, meaning:
1. taking an AskScript program and environment
2. parsing it
3. transpiling to AskCode
4. executing AskCode
5. returning the result
as opposed to just parsing AskScript or just executing AskCode.
Probably we should rename it. Feel free to create an issue to this and discuss the new name there.

RichardM99:
@czerwinskilukasz1 

I'm afraid I don't have permission to merge.

src/askscript/__tests__/90-misc/fetch.test.ts
```diff
@@ -0,0 +1,41 @@
+jest.mock('node-fetch');
```
mhagmajer, line 1: :+1:

czerwinskilukasz1:
> @czerwinskilukasz1
    >
    > I'm afraid I don't have permission to merge.
    
    Ouch, forgot about permission. I am doing the merge now, congrats for making your first change in AskQL repository! <U+1F680>
    How did you like it? ;)

RichardM99:
@czerwinskilukasz1 

Thanks! It was great - thanks again for the support.

Will hopefully contribute again soon.

czerwinskilukasz1:
@RichardM99 , sounds great!  If you feel you would like any suggestion on which issue to choose, just tag me.

RichardM99:
@czerwinskilukasz1 

If you would like to suggest an issue I will gladly work on it.

czerwinskilukasz1:
Sounds great, @RichardM99  🙂 
How do you like this one? https://github.com/xFAANG/askql/issues/269 - it's about changing a way a resource (a Typescript function) works.
There is also https://github.com/xFAANG/askql/issues/181 , which is about adding _some_ real code examples in AskScript language.

We scheduled these issues (and a couple more) for [milestone 1.3](https://github.com/xFAANG/askql/milestone/3), which we would like to release soon.

RichardM99:
@czerwinskilukasz1 

I can take https://github.com/xFAANG/askql/issues/269


BTW - I think when this PR was merged the commits were not squashed.

czerwinskilukasz1:
Sounds great!

PR #383, opened 2020-07-09
fix(askscript): fix tricky array expression

### LINKED ISSUE #382: [askscript] Array value gets merged with the previous line
Consider the following program:
```
ask { 
    'abcd'
    [1]
} 
```

The expected return value is `[1]`.
The actual return value is `b`. 
Why? Because we treat newlines like any other whitespace, so
```
'abcd'
    [1]
```
is seen the same as 
``` 'abcd' [1] ```
and
``` 'abcd'[1] ```, which is obviously 'b'.

czerwinskilukasz1:
Closes #382


mhagmajer:
I like the example program :).

Pending discussion on https://github.com/xFAANG/askql/issues/382

mhagmajer:
@czerwinskilukasz1 can you explain how this is solved and what are the consequences of this change to the parser?

PR #393, opened 2020-07-15
feat: min Closes #369

### LINKED ISSUE #369: Implement 'min' resource
There is [`max`](https://github.com/xFAANG/askql/blob/master/src/askvm/resources/math/max.ts) in Ask, but there is no `min`.

Hint: use `max` ([link](https://github.com/xFAANG/askql/blob/master/src/askvm/resources/math/max.ts)) as an example for `min`.

199201shubhamsahu:
Passes all tests on running `npm test`. Couldn't figure the commented test case in the max.test.ts file which is marked as TODO. Hence, replicated the similar case in the min.test.ts as well and kept it commented and marked as TODO

### DIFF
--- a/src/askvm/resources/math/__tests__/min.test.ts
+++ b/src/askvm/resources/math/__tests__/min.test.ts
@@ -1,0 +1,35 @@
+import { runUntyped } from '../../..';
+import { parse } from '../../../../askcode';
+import * as core from '../../core';
+import { min } from '../min';
+
+function ask(code: string) {
+  return runUntyped(
+    {
+      values: {
+        obj: {
+          a: 4,
+          b: 123,
+          c: 345,
+        },
+        scores: [1, 2, 3, 953, 5],
+      },
+      resources: { ...core, min },
+    },
+    parse(code)
+  );
+}
+
+test('min', async () => {
+  await expect(ask('min(1, 2)')).resolves.toBe(1);
+  // await expect(() => ask(`min(1, '2')`)).rejects.toBeTruthy(); // TODO FIXME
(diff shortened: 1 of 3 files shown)


src/askvm/resources/math/__tests__/min.test.ts
```diff
@@ -0,0 +1,35 @@
...
+      },
+      resources: { ...core, min },
+    },
+    parse(code)
+  );
+}
+
+test('min', async () => {
+  await expect(ask('min(1, 2)')).resolves.toBe(1);
+  // await expect(() => ask(`min(1, '2')`)).rejects.toBeTruthy(); // TODO FIXME
```
mhagmajer, line 25: This we can uncomment once we have type checks

src/askvm/resources/math/__tests__/min.test.ts
```diff
@@ -0,0 +1,35 @@
...
+  await expect(ask('min(2, 3, min(4, min([1, 13], 5)))')).resolves.toBe(1);
+  await expect(ask('min([2, 3, [1, 6]])')).resolves.toBe(1);
+  await expect(ask('min(scores)')).resolves.toBe(1);
+  await expect(ask(`object('a', 2, 'b', 4)`)).resolves.toStrictEqual({
+    a: 2,
+    b: 4,
+  });
+  await expect(ask(`min(object('a', 2, 'b', 4))`)).resolves.toBe(2);
```
mhagmajer, line 33: For now it's okay, but eventually (with type checks) these are going to throw since `'a'` and `2` have different types and can't be compared to find the smaller one

Issue #390, opened 2020-07-13
whitespace handling bug in if() 


MartaBBRM:
Dear First Time Contributor, 

This ticket evolved. Please scroll through the comments to find out what was discovered.
In case of any questions, feel free to tag me (@czerwinskilukasz1) and type your question. 

Cheers, 
Łukasz 
Core AskQL Contributor 

--------------

```
ask { let sumOfSquares = fun (n) {
    if (n :lessThanOrEquals (0) ) {
        return 0
        
    }
    (n*n) + sumOfSquares(n-1)
}
    // sum of squares
    // 4*4 + 3*3 + 2*2 +1*1
    //3*3 + 2*2 + 1*1
    //1*1
    // 0*0
    
    sumOfSquares(-4)
}
```
it happened in the cli (AskVM v. 1.2.0) and it was expected to run ;)
The error massage:

```
SyntaxError: Expected " ", ".", "//", ":", ";", "[", "\n", "\r", "\t", "}", [/], or [\-<>+*\^%=&|] but "{" found.
Location: {
  "start": {
    "offset": 69,
    "line": 2,
    "column": 35
  },
  "end": {
    "offset": 70,
    "line": 2,
    "column": 36
  }
}
```

MartaBBRM:
@czerwinskilukasz1

czerwinskilukasz1:
Thanks.
It seems that the underlying issue is whitespace handling before the trailing bracket in `if ()`

The minimal example that does not work:
```
ask {
    if(3 :lessThan (4) ) {
        return 1
    }
} 
```

The culprit is the space in `4) )`.

czerwinskilukasz1:
Taken by @199201shubhamsahu

czerwinskilukasz1:
(For some reason I cannot add you as an assignee :o, so I listed you in a comment)

PR #394, opened 2020-07-15
fix(askscript): fix whitespace issue in control flows (#390)

199201shubhamsahu:
The issue is not only in `if` but also in `while`, `forOf`, and `forIn` control flows. Hence, modified the code parser for all the required control flows. Passes all required tests.


199201shubhamsahu:
@czerwinskilukasz1 I have added the .ask files in https://github.com/xFAANG/askql/tree/master/src/askscript/__tests__/04-control-flow in the following commit in PR. Please review and let me know if anything needs to be added, removed or modified.

czerwinskilukasz1:
LGTM, thanks, Shubham :)

PR #406, opened 2020-07-17
fix(askscript): make if/while/... without curly braces to throw Syntax Error

### LINKED ISSUE #403: if block without braces silently executes wrong branch
Consider the following AskScript program:
```
ask {
  const n = 3
  if ((n==0) || (n == 1))
    return true
  
  return false
}
```

The expected result is: Syntax error because the if block should have braces.
The actual result is: `true`

czerwinskilukasz1:
Closes #403


czerwinskilukasz1:
Hi @mhagmajer , would you have a second to check this PR please?

src/askscript/parser/askscript.grammar.pegjs
```diff
@@ -104,7 +104,7 @@ variableDefinition_type = ws* ':' ws* t:type { return t }
 
 // === value ===
 
-value = e:nonArithmExpression opVals:operValue* {  assertAllOperatorsMatch(opVals); return new ask.Value(e, opVals) }
+value = !(('const' / 'let' / 'if' / 'while' / 'for' / 'return' ) (ws / eof)) e:nonArithmExpression opVals:operValue* {  assertAllOperatorsMatch(opVals); return new ask.Value(e, opVals) }
```
mhagmajer, line 107: These could be defined separately as keywords, this list may grow

PR #408, opened 2020-07-19
feat(askvm): toInt

RichardM99:
- Will resolve https://github.com/xFAANG/askql/issues/269#issuecomment-660027793

@czerwinskilukasz1 

Main issues I have;

- Passing `true`, `false` or `null` to `toInt` results in the arg being converted into an instance of `AskCode` e.g. `toInt(false)` --> `false` becomes `AskCode { name: false, params: undefined }` inside of `toInt`
- I assume I should not have to do something like

```
let val;
if (value instanceof AskCode) {
     val = value.name;
}
return Math.round(val);
```

- I guess there is some step I am missing - if you could advise would be great!
- Also not sure if `toInt(object('a', 1))` is the correct way to test `toInt({a: 1})`

### DIFF
--- a/src/askvm/resources/math/__tests__/toInt.test.ts
+++ b/src/askvm/resources/math/__tests__/toInt.test.ts
@@ -1,0 +1,103 @@
+import { runUntyped } from '../../..';
+import { parse } from '../../../../askcode';
+import * as core from '../../core';
+import { toInt } from '../toInt';
+
+describe('math - toInt', () => {
+  let result: any;
+
+  it(`should return 2 when given 2.3`, async () => {
+    result = await ask(`toInt(2.3)`);
+
+    expect(result).toBe(2);
+  });
+
+  it(`should return 2 when given '2.3'`, async () => {
+    result = await ask(`toInt('2.3')`);
+
+    expect(result).toBe(2);
+  });
+
+  it(`should return 3 when given 2.5`, async () => {
+    result = await ask(`toInt(2.5)`);
+
+    expect(result).toBe(3);
+  });
+
+  it(`should return 3 when given '2.5'`, async () => {
+    result = await ask(`toInt('2.5')`);
+
+    expect(result).toBe(3);
+  });
+
+  it(`should return -2 when given -2.5`, async () => {
+    result = await ask(`toInt(-2.5)`);
+
+    expect(result).toBe(-2);
+  });
+
+  it(`should return -2 when given '-2.5'`, async () => {
+    result = await ask(`toInt('-2.5')`);
+
+    expect(result).toBe(-2);
+  });
+
+  it(`should return -3 when given -2.51`, async () => {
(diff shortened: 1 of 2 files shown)


czerwinskilukasz1:
Hi @RichardM99 ,

Thanks for working on the issue!
Let me answer in order:

> * Will resolve [#269 (comment)](https://github.com/xFAANG/askql/issues/269#issuecomment-660027793)

Great that you added issue # to the description.
If the PR is not ready yet, it's good to convert it to draft to indicate it's not ready for review yet:
image


> Main issues I have;
> 
> * Passing `true`, `false` or `null` to `toInt` results in the arg being converted into an instance of `AskCode` e.g. `toInt(false)` --> `false` becomes `AskCode { name: false, params: undefined }` inside of `toInt`
> * I assume I should not have to do something like
> 
> ```
> let val;
> if (value instanceof AskCode) {
>      val = value.name;
> }
> return Math.round(val);
> ```
> 
> * I guess there is some step I am missing - if you could advise would be great!

These are great questions.  
AskVM was written by @mhagmajer , he will be the best person to answer your questions.  I suspect it might be a bug, since numbers are passed as plain Javascript values.

> * Also not sure if `toInt(object('a', 1))` is the correct way to test `toInt({a: 1})`

I think so, but let's wait for @mhagmajer to confirm.

czerwinskilukasz1:
@mhagmajer , would you have a sec to answer Richard's questions?

mhagmajer:
@RichardM99 I see that you use `parse` for AskCode which is more low level than `parse` for AskScript. Let's look at this example from CLI:

```
🦄 true
boolean ask(call(get('true')))
true
```

Expression `true` in AskScript is translated to `ask(call(get('true')))` in AskCode. As you can see, in AskCode `true` is actually a function than needs to be called to get the value for `true`. There reason we did it like so is that there is no constants in AskCode - everything is a resource that needs to be called to get value.

https://github.com/xFAANG/askql/blob/master/src/askvm/resources/core/true.ts
https://github.com/xFAANG/askql/blob/master/src/askvm/resources/core/false.ts
https://github.com/xFAANG/askql/blob/master/src/askvm/resources/core/empty.ts

Hope this helps. If you have any more questions please let us know :)

RichardM99:
@mhagmajer 

Thanks for response! I updated the tests.

Don't really like how for the error tests I am still using the `ask` function rather than the `execution` function -  but using the `execution` function when checking if the correct error is being thrown throws something like

`SyntaxError: Expected " ", ")", ",", ".", ":", "[", "\n", "\r", "\t", [/], [0-9], or [\-!<>+*\^%=&|] but "a" found.`

rather than the expected error (but works as expected when using `ask`)

RichardM99:
@czerwinskilukasz1 

Done.

RichardM99:
@czerwinskilukasz1 

Please feel free to suggest another issue I could possibly work on!

PR #413, opened 2020-07-26
feat(askvm): add `hasKey` resource

### LINKED ISSUE #400: Implement `hasKey` resource
I wanted to write a piece of code that in Javascript looks like this:
```
if (!(a in leftMostPosition)) {
   leftMostPosition[a] = i;
}
```
Alas, I couldn't find any way to express it, because I didn't find a resource for it.
Please note that comparing `leftMostPosition[a]` with any value would not work as there is no `undefined` value in AskScript, only `null`, which could be a legitimate value of `leftMostPosition[a]`.

cpelican:
Closes #400


src/askscript/__tests__/07-arrays/program17e-object_has_key_is_false.ask
```diff
@@ -0,0 +1,9 @@
+ask {
+  const countries = {usa: 'USA', spain: 'Spain', russia: 'Russia'};
+  const sweden = 'Sweden';
+  if(!(countries:hasKey(sweden:toLowerCase))) {
```
czerwinskilukasz1, line 4: I would aim for creating a minimal working tests, so I'd remove `const sweden = 'Sweden';` and replaced `sweden:toLowerCase` with just `'sweden'`.

src/askvm/resources/list/hasKey.ts
```diff
@@ -0,0 +1,11 @@
+import { any, resource } from '../../lib';
+
+export const hasKey = resource({
+  type: any,
+  async resolver(listOrObject: object, key: any): Promise<boolean> {
+    if (listOrObject == null || typeof listOrObject !== 'object') {
```
czerwinskilukasz1, line 6: Great check! :)

src/askscript/__tests__/07-arrays/program17e-object_has_key_is_false.ask
czerwinskilukasz1: But it's fine with the code you wrote.

Issue #273, opened 2020-06-27
Implement toFloat


czerwinskilukasz1:
Following [@mhagmajer's comment](https://github.com/xFAANG/askql/issues/269#issuecomment-650568501), a `toFloat` resource is needed, which converts the argument to float.

Hint for new contributors:
Check how `toInt` is implemented and write `toFloat` in a similar manner.

If you have any questions on how to start, please write a comment with a tag: @czerwinskilukasz1 or @mhagmajer, or join our [Discord Community](https://discord.gg/pYdzypH) and let us know there. We will be very happy to help you start! :)

Cheers,
Łukasz (@czerwinskilukasz1)
AskQL Core Developer

engylemure:
@mhagmajer and @czerwinskilukasz1  Considering that the JS does not have an internal difference between integer and float since they both are [Number](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number) this seems to be tricky, are you guys thinking about implementing wrappers for more explicitly handling the differences between the values considering casts?

czerwinskilukasz1:
Hi @engylemure , great question and sorry for taking a bit longer to reply, I was busy with solving tickets, haha :) 
We are going to implement type checking (per https://github.com/xFAANG/askql/issues/45) and `float` and `int` are separate types.
We already have them listed [in our grammar](https://github.com/xFAANG/askql/blob/master/src/askscript/parser/askscript.grammar.pegjs#L245) as well as other built-in types.
Does this answer your question?

jamesd35:
Hi can I work on this feature?

czerwinskilukasz1:
Sure, go ahead @jamesd35 !

jamesd35:
Hi, I have a question on the intended behavior for integers passed into ToFloat.

Based on the float definition it seems that any integers will not be validated
`export const float = type<Type<number>, Type<number>>(number, {
  name: 'float',
  validate: (value): value is number => value % 1 !== 0,
});`

Currently if I pass in an integer, it returns it as an int
`toFloat(4)
int ask(call(get('toFloat'),4))
4`

Is this correct or should it return a float such as 4.0?

czerwinskilukasz1:
Good catch, jamesd35.  Since @mhagmajer wrote the types and validators, let's wait for his reply.

mhagmajer:
[...]

PR #439, opened 2020-08-09
feat(askvm): toFloat

jamesd35:
Feature "Implement toFloat #273 "

Note that passing in an integer value will result in int data type due to the way int and float are defined.


mhagmajer (approved):
Great job @jamesd35! Thanks for submitting a PR for this! 🚀

PR #459, opened 2020-08-19
feat(askscript): allow `else if` construct

### LINKED ISSUE #378: Allow 'else if' construct
Currently we don't have `switch` nor we permit `else if`, because `else` expects a block with curly braces.
How about we permit `else if`, such as:
```
    if (len == 0) {
      return 'no one likes it';
    } else if (len == 1) {
      return concat(names[0], ' likes it');
    } else if (len == 2) {
      return concat(names[0], ' and ', names[1], ' like it');
    } else if (len == 3) {
      return concat(names[0], ', ', names[1], ' and ', names[2], ' like it');
    } else {
      return concat(names[0], ', ', names[1], ' and ', len, ' others like it');
    }
```

markkulube:
Closes #378: Allow 'else if' construct


src/askscript-vscode/package.json
```diff
@@ -1,25 +1,34 @@
 {
-    "name": "askscript-language",
```
czerwinskilukasz1, line 2: This file is heavily changed for no good reason. 
I think it's just reformatting. For readability, how about you revert the changes in this file?

src/askscript-vscode/vsc-extension-quickstart.md
```diff
@@ -2,28 +2,28 @@
 
 ## What's in the folder
 
-* This folder contains all of the files necessary for your extension.
```
czerwinskilukasz1, line 5: Any good reason for these formatting changes? I would like to avoid you changing from * to - and then somebody else changing back from - to *, so if reasonable, please revert changes in this file.

src/askscript/__tests__/04-control-flow/program10d_2b-if_else_if_else.ast.tsx
```diff
@@ -0,0 +1,46 @@
+export = (
```
czerwinskilukasz1, line 1: I think there should be much more files added than just one .tsx file.
Look what I got after running `npm run build && npm test`:

image


All these files should be added to your commit.


Also, your changes to the code resulted in changes in the existing .ast.tsx files. You should commit them too:
image

czerwinskilukasz1 (changes requested):
Hi Mark, 
Great to see your PR, congrats!
There is just a couple of details to fix. 

Most importantly, it seems to me that *.formatted.ask files are not just reformatted .ask files. They are missing their `else` and `else if` branches, e.g. file program10d_3d-if_else_if_else.ask.formatted.ask is:
```
ask() {
  let len = 10
  if ((get('len')==12)) {
    return '3DF'
  }
}
```

while it should be a reformatted version of `program10d_3d-if_else_if_else.ask` file, which is:
```
ask {
    let len = 10
    if (len == 12) {
        return '3DF';
    } else if (len == 14) {
        return '3DF';
    } else if (len == 16) {
        return '3DF';
    } else {
        return '3DP';
    }
}
```

src/askscript-vscode/package.json
markkulube: Yes, it's only a reformatting issue. I executed `npm run prettier`, and inadvertently reformatted reformatted files that were not to intended to be.

**Previous formatting has been restored.**

src/askscript-vscode/vsc-extension-quickstart.md
markkulube: Yes, it's only a reformatting issue. I executed `npm run prettier`, and inadvertently reformatted reformatted files that were not to intended to be.

**Previous formatting has been restored.**

src/askscript/__tests__/04-control-flow/program10d_2b-if_else_if_else.ast.tsx
markkulube: Indeed, all missing *.askcode and *.ast.tsx have been pushed.

I am using git client sourcetree which does makes it easier identify, stage, and commit untracked files.

Let me know if anything is still missing.

markkulube:
@czerwinskilukasz1 the main issue has been resolved.

It was great putting theory into practice. 

Since an if-else block amounts to a linked list (If->ElseIf->Elseif->...->Else), we have to traverse the prettier-plugin-askscript/print.ts/node to access inner/subsequent blocks or branches. It's the cleanest way I could do it.

Thus we can generate the correct and complete *.askcode and *.ask.formatted.ask files on executing the test scripts.

Non of the existing *.askcode have been changed, which tells the `else if else if` implementation is yet affects behavior of the current if and if-else clauses.

Still do let me know if anything is amiss.

markkulube:
@czerwinskilukasz1 just a couple notes:
- Another test for the case when the `if else if else` ladder terminates with an `else if` branch. 
- In reality, since we reusing the If jsx structure, the behavior is the same as when the ladder only constitutes an `if` clause.
`
Welcome to AskQL CLI!

For multi-line mode please type:
     .editor
🦄 .editor
// Entering editor mode (^D to finish, ^C to cancel)
let len = 333
if(len == 3) {
return '3A FAILED';
} else if(len == 10) {
return '3A FAILED';
} else if(len == 333) {
return '3A PASSED';
}

string ask(let('len',333),if(call(get('=='),get('len'),3),block(return('3A FAILED')),block(if(call(get('=='),get('len'),10),block(return('3A FAILED')),block(if(call(get('=='),get('len'),333),block(return('3A PASSED')),block()))))))
'3A PASSED'
🦄              
`

czerwinskilukasz1 (approved):
LGTM

PR #469, opened 2020-09-10
feat(operator): div operator

shortcircu1t:


mhagmajer (approved):
Great job @shortcircu1t! Thanks for sending a PR and adding tests. 🙌 

For modulo we have `%` assigned so you can do `5 % 3` (gives 2) similarly like JS, but no operator for div. Maybe one day we can revisit this.

Issue #483, opened 2020-09-14
How to concatenate arrays?


mhagmajer:
Looks like we need something for extending arrays

YonatanKra:
@mhagmajer 
Can we call it `extend`?
Should it also extend objects?

PR #492, opened 2020-10-07
feat(list): add extend resource

### LINKED ISSUE #257: Implement extend resource create a new object with added property
While we have a resource for setting a value in an array (`set` resource), I think we don't have such for objects.

It is up to discussion whether a new resource should be created or rather `set` should be extended to work for objects.

YonatanKra:
closes #257 #483


mhagmajer (changes requested):
Good idea!

Let's:

1. Add the new resource(s) in `../index.ts` so they are actually exported
2. Add some unit tests on AskScript level (eg. src/askscript/__tests__/07-arrays)
3. Split extending arrays and objects into two separate resources - for instance `listExtend` and `objectExtend` - it will be much easier to add typing layer if these have the same input and output types

src/askvm/resources/list/arrayExtend.ts
```diff
@@ -0,0 +1,25 @@
...
+  throw new Error('Expecting 2 arrays in arrayExtend');
+};
+
+function asyncExtendArray(list: any[], listExtension: any[]) {
+  return [...list, ...listExtension];
+}
+
+export const arrayExtend = resource<any, any[]>({
+  type: any,
+  async resolver(list: any[] | {}, listExtension: any[] | {}): Promise<any> {
```
mhagmajer, line 13: rm `| {}`

src/askvm/resources/list/objectExtend.ts
```diff
@@ -0,0 +1,21 @@
+import { any, resource } from '../../lib';
+
+const typesError = () => {
+  throw new Error('Expecting 2 objects in objectExtend');
+};
+
+export const objectExtend = resource<any, any[]>({
+  type: any,
+  async resolver(list: any[] | {}, listExtension: any[] | {}): Promise<any> {
```
mhagmajer, line 9: rm `any[]`

src/askvm/resources/list/index.ts
```diff
@@ -2,3 +2,5 @@ export { find } from './find';
 export { hasKey } from './hasKey';
 export { length } from './length';
 export { set } from './set';
+export { objectExtend } from './objectExtend';
```
mhagmajer, line 5: Can we move that under a different directory?

src/askvm/resources/list/arrayExtend.ts
```diff
@@ -0,0 +1,25 @@
+import { any, resource } from '../../lib';
+
+const typesError = () => {
+  throw new Error('Expecting 2 arrays in arrayExtend');
+};
+
+function asyncExtendArray(list: any[], listExtension: any[]) {
+  return [...list, ...listExtension];
+}
```
mhagmajer, line 9: this is not actually async, let's simplify with `Array.prototype.concat` maybe?

mhagmajer (approved):
🚀 😎

PR #505, opened 2020-10-09
Return the string of NaN and Infinity in the playground

### LINKED ISSUE #367: [playground] Infinity is reported as null
Consider the following program:

```
ask {
  const factorial: int(int) = fun(n:int):int {
    if (n < 2) {
      return n
    }

    n * factorial(n - 1)
  }

  200:factorial
}
```

In CLI the result is:
```
float ask(const('factorial',fun(let('n',get('$0')),if(call(get('<'),get('n'),2),block(return(get('n'))),block()),call(get('*'),get('n'),call(get('factorial'),call(get('-'),get('n'),1))))),call(get('factorial'),200))
Infinity
🦄 
```

However, [Playground](https://cli.askql.org/) displays:
```
null
```

YonatanKra:
Closes #367 
See my comment in #367 regarding Infinity and NaN in JSON

### DIFF
--- a/src/playground-backend/express/demoAskScriptServer.ts
+++ b/src/playground-backend/express/demoAskScriptServer.ts
@@ -36,7 +36,21 @@ app.get('/version', async (req, res) => {
   res.json({ version: packageInfo.version });
 });
 
+function logError(id: string, code: string, e: Error) {
+  console.error(id + ' -- ' + new Date().toString());
+  console.error(id + ' -- ' + code);
+  console.error(id + ' -- ' + e);
+  console.error('\n\n');
+}
+
 app.post('/askscript', async (req, res) => {
+  function reportError(e: Error, message: string) {
+    res.status(400).json({
+      message,
+      error: e.toString(),
+    });
+  }
+
   const code: AskScriptCode = req.body.code;
 
   if (typeof code !== 'string') {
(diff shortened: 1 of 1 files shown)


mhagmajer:
@YonatanKra thanks for creating another PR 💪

Can you explain your rationale of fixing this and how this PR was tested? As you have noticed in https://github.com/xFAANG/askql/issues/367#issuecomment-705983423 neither `Infinity` nor `NaN` can be transferred in JSON and playground is based on sending an AskQL query and receiving a JSON response so I'm not sure how this can be solved without challenging that assumption.

YonatanKra:
> @YonatanKra thanks for creating another PR 💪
> 
> Can you explain your rationale of fixing this and how this PR was tested? As you have noticed in https://github.com/xFAANG/askql/issues/367#issuecomment-705983423 neither `Infinity` nor `NaN` can be transferred in JSON and playground is based on sending an AskQL query and receiving a JSON response so I'm not sure how this can be solved without challenging that assumption.

What I did was change how the playground works. It returns a string of Infinity or NaN in case these values are returned (instead of returning a json.stringifyof them.

Tested on local playground

src/playground-backend/lib/utils.ts
```diff
@@ -15,3 +19,31 @@ export const sendJson = (
...
+  let result = await runUntyped(baseEnvironment, askCode, []);
+  if (
+    result === Infinity ||
+    result === -Infinity ||
+    (typeof result === 'number' && isNaN(result))
+  ) {
+    result = result.toString();
+  }
+  return result;
+}
```
YonatanKra, line 49: @mhagmajer this is the "magic". Would it be valid to add this logic to the middleware as well?

mhagmajer:
Let's wait for the https://github.com/xFAANG/askql/issues/367#issuecomment-706162367 to resolve

mhagmajer (changes requested):
Let's wait for discussion at https://github.com/xFAANG/askql/issues/367#issuecomment-706162367 to resolve

czerwinskilukasz1:
@YonatanKra , I believe we should first finish our discussion in #367 and only after that create any PRs. 
For now I would pause this PR until we come up to a conclusion how to solve it.

YonatanKra:
> @YonatanKra , I believe we should first finish our discussion in #367 and only after that create any PRs.
> For now I would pause this PR until we come up to a conclusion how to solve it.

Drafted :)

Issue #312, opened 2020-07-02
Object cannot be used as value in query?


mhagmajer:
<img width="1045" alt="Screenshot 2020-07-02 at 12 55 11" src="https://user-images.githubusercontent.com/2731042/86350703-6164c680-bc63-11ea-8426-4eb16ff46848.png">

```
ask {
  query {
    firstName1: { a:1 }
    firstName2: ({ a:1 })
  }
}
```

Returned object is empty, but `{ "a": 1 }` was expected

mhagmajer:
@czerwinskilukasz1 what do you think about it? Does it make sense to mix the query and value semantics like this? Where should we place a border between `query` and regular value expression?

czerwinskilukasz1:
I am for allowing this kind of syntax.  Why do we permit strings or ints but would forbid dictionaries?
@mhagmajer , did you do research why it doesn't work? Is it parser or askcode/askvm?

YonatanKra:
Adding this test case here:
```
// TODO :: should we make this work?
  // it(`should return an object`, async function () {
  //   const expectedObjectValue = {
  //     objectReturned: values.objectValue,
  //   };
  //
  //   const queryForAnObjectValue = `query(node('objectReturned',f(object('a',call(get('get'),'objectValue')))))`;
  //
  //   await expect(ask(queryForAnObjectValue)).resolves.toStrictEqual(
  //     expectedObjectValue
  //   );
  // });
```

PR #506, opened 2020-10-09
test(query): added query unit tests

YonatanKra:
note the TODO is a test we should decide if we want it or not (#312 )

### DIFF
--- a/src/askvm/resources/core/__test__/query.test.ts
+++ b/src/askvm/resources/core/__test__/query.test.ts
@@ -1,0 +1,93 @@
+import { runUntyped } from '../../..';
+import { parse } from '../../../../askcode';
+import * as core from '../../core';
+import { query } from '../../core';
+
+const values = {
+  objectValue: {
+    x: [1, 2, 3, 4],
+  },
+  stringValue: 'string',
+  numberValue: 5,
+  booleanValue: true,
+  arrayVal: [1, 2, 3, 4],
+};
+
+function ask(code: string) {
+  return runUntyped(
+    {
+      resources: { ...core, query },
+      values,
+    },
+    parse(code)
+  );
+}
+
+describe('hasKey', () => {
+  it(`should return an array value`, async function () {
+    const expectedValueForArray = {
+      objectReturned: values.arrayVal,
+    };
+
+    const inLineArray = [1, 2, 3, 4];
+    const expectedValueForInlineArray = {
+      objectReturned: inLineArray,
+    };
+
+    const queryForAnArrayValue = `ask(query(node('objectReturned',f(call(get('get'),'arrayVal')))))`;
+    const queryForInlineArray = `ask(query(node('objectReturned',f(list(${inLineArray.join(
+      ','
+    )})))))`;
+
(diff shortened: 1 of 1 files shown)


src/askvm/resources/core/__test__/query.test.ts
```diff
@@ -0,0 +1,93 @@
+import { runUntyped } from '../../..';
+import { parse } from '../../../../askcode';
```
mhagmajer, line 2: I wonder if for resource unit test we should be calling the resource directly or keep going through the AskCode parser like you're doing here 🤔 . Maybe it would be easier to catch any potential problems and check for different scenarios if we were calling `query` resource directly. What do you guys think @YonatanKra @czerwinskilukasz1?

YonatanKra: I didn't even notice it. Just my IDE auto import doing its thing.
Definitely better to do direct here. The post build tests willmake sure it is available in the barrel file.

YonatanKra: Looking closer at your response I now see you mean to test the function directly. 
Should make for easier testing.

I'll try that.

I just took existing tests and followed conventions in this area.

Both ways work for me as long as I can test. 

Main point is to decide on good conventions and stick to them so read8ng one part of code would look similar to others (in regards to style etc.).

mhagmajer: @YonatanKra yes, you are right that we're missing consistency there. I think personally that low level unit test for a function should just test this function and not rely to much on other code - this way they are more resilient and allow for testing a greater variety of edge cases. We also have this other kind of tests that are more like "unit integration tests" where we checking results for askScript programs.

Either way both approaches here are acceptable.

src/askvm/resources/core/__test__/query.test.ts
```diff
@@ -0,0 +1,93 @@
...
+  //   const expectedObjectValue = {
+  //     objectReturned: values.objectValue,
+  //   };
+  //
+  //   const queryForAnObjectValue = `query(node('objectReturned',f(object('a',call(get('get'),'objectValue')))))`;
+  //
+  //   await expect(ask(queryForAnObjectValue)).resolves.toStrictEqual(
+  //     expectedObjectValue
+  //   );
+  // });
```
mhagmajer, line 92: Instead of having commented code here can we link to a GitHub issue in a comment and maybe have that code there?

mhagmajer (commented):
@YonatanKra almost there! Please add a test case of calling `query` resolver directly if you think that would be valuable

src/askvm/resources/core/__test__/query.test.ts
czerwinskilukasz1: @YonatanKra I am fine with leaving the code there, but please please add a link to the GitHub issue.

YonatanKra: I don't like comments in code - especially when we have git :)
So I've copied the code to the issue itself.

mhagmajer (approved):
👏

Issue #387, opened 2020-07-09
[askscript] How about we implement push/append to an array?


czerwinskilukasz1:
While writing AskScript examples, I wanted to append to an array twice already.
Alas, the best I could do is to write:
```
arr:set(arr:length, value)
```
which is not that easy readable.

I'd prefer to have push/append, like in Javascript:
```
  let result = [];
  for (let i = 1; i <= m; ++i) {
    result.push(n*i);
  }
```

mhagmajer:
@czerwinskilukasz1 `push` is agains the principle of immutable data. If we come up with some syntax sugar it will still translate to:

```
result = result:push(n*i)
```

czerwinskilukasz1:
Sure sure, I'm totally fine with `result = result:push(elem)`.

avats-dev:
Hey @mhagmajer  and @czerwinskilukasz1 , as I have asked elsewhere also, if you could guide me where and what needs to be done to get this fix, I would do so.
> I'm new to this codebase, but  want to contribute.

czerwinskilukasz1:
Hi Aditya, welcome to AskQL again :) 
This task is pretty straightforward, comparing to the one with double `let` expression. The task is to create a new _resource_ (aka function) called `push`, which will live in `src/askvm/resources/list/`, same as `set` resource, which is currently defined.
Please use `src/askvm/resources/list/set.ts` as an example how resources are defined and how they work.
Let me know if you have more questions.

czerwinskilukasz1:
Hi Aditya, welcome to AskQL again :) 
This task is pretty straightforward, comparing to the one with double `let` expression. The task is to create a new _resource_ (aka function) called `push`, which will live in `src/askvm/resources/list/`, same as `set` resource, which is currently defined.
Please use `src/askvm/resources/list/set.ts` as an example how resources are defined and how they work.
Let me know if you have more questions.  @avats-dev

mhagmajer:
Closed with https://github.com/xFAANG/askql/pull/510 from @bardeutsch 👏

PR #510, opened 2020-10-10
feat(array): added push for array

bardeutsch:
filling request in issue #387


src/askvm/resources/list/push.ts
```diff
@@ -0,0 +1,9 @@
+import { any, resource } from '../../lib';
+
+export const push = resource({
+  type: any,
+  async resolver(arr: Array<any>, value: any) {
+    arr.push(value);
```
YonatanKra, line 6: @mhagmajer - are we still immutable with this?

YonatanKra:
@bardeutsch thanks for the contribution.
One thing I'm missing is a post build test in `./src/askScript/__tests__/arrays`
I've also left a comment for @mhagmajer.

src/askvm/resources/list/push.ts
mhagmajer: Nope, we need to create a new array here. Also, it would be good to add a unit test that verifies immutability.

bardeutsch:
@YonatanKra I added the test, I hope it's what you meant.
@mhagmajer  where should I add a test like that? Do you have something similar?

mhagmajer:
Place @YonatanKra is good for this kind of test. We would need to check that the new `push` resource doesn't modify the original array

YonatanKra:
> @YonatanKra I added the test, I hope it's what you meant.
> @mhagmajer where should I add a test like that? Do you have something similar?

@bardeutsch 
Thanks. Your sanity test is fine :)
Regarding immutability - check what I did here:
https://github.com/xFAANG/askql/pull/511/files#diff-a8111cb4e1a0e3bb029f4dda4e886973
I think it is good enough for the `push` case (just change `map` with `push`).

bardeutsch:
> > @YonatanKra I added the test, I hope it's what you meant.
> > @mhagmajer where should I add a test like that? Do you have something similar?
> 
> @bardeutsch
> Thanks. Your sanity test is fine :)
> Regarding immutability - check what I did here:
> https://github.com/xFAANG/askql/pull/511/files#diff-a8111cb4e1a0e3bb029f4dda4e886973
> I think it is good enough for the `push` case (just change `map` with `push`).

Cool, done :)

YonatanKra:
LGTM
@mhagmajer ? Any comments?

mhagmajer (approved):
Nice work @bardeutsch! 👍

PR #511, opened 2020-10-11
test: add immutability tests for map

YonatanKra:
Plus a small cleanup of `each`

### DIFF
--- a/src/askvm/resources/list/__tests__/map.test.ts
+++ b/src/askvm/resources/list/__tests__/map.test.ts
@@ -3,7 +3,6 @@ import { parse } from '../../../../askcode';
 import * as core from '../../core';
 import { map } from '../map';
 import { concat } from '../../string';
-import { each } from '../each';
 
 const values = {
   clientNames: ['A', 'B', 'C', 'D', 'E', 'F', 'G'],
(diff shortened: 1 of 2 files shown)


src/askvm/resources/list/__tests__/map.test.ts
```diff
@@ -44,3 +44,11 @@ test('map', async () => {
     )
   ).resolves.toStrictEqual(namesWithPrefix);
 });
+
+test('map should not mutate the array', async () => {
+  const array = [1, 2, 3, 4];
+
+  const anotherArray = await map.resolver(array, (item: number) => item + 1);
+
+  expect(array !== anotherArray).toEqual(true);
```
mhagmajer, line 53: Let's make use of built-in assertion methods

```suggestion
expect(array).not.toBe(anotherArray);
```

Can you also add a check to see if the array was mapped correctly and that the previous array is left intact?

mhagmajer (changes requested):
Nice @YonatanKra! 👌

mhagmajer:
Decided to merge it like this :+1:

mhagmajer (approved):
:+1:

PR #513, opened 2020-10-11
feat(object): object string property

### LINKED ISSUE #512: Enable string property names
```
ask {
	const starWarsCharacters = fetch('https://swapi.dev/api/people'):at('results')
  starWarsCharacters:map(fun(character) {
    {
      'Character name': character:at('name'),
      'Hair Color': character:at('hair_color'),
      'Year of Birth': character:at('birth_year'),
      gender: character:at('gender'),
    }
  })
}
```

YonatanKra:
Closes #512 
Edit: Has issue with post build tests. Works on playground.


src/askscript/parser/askscript.grammar.pegjs
```diff
@@ -270,6 +270,7 @@ mapEntryList =
 
 mapEntry = 
     i:identifier ws* ':' ws* v:value {                 return new ask.MapEntry(i, v) }
+    / ws* i:string ws* ':' ws* v:value {                 return new ask.MapEntry(i, v) }
```
czerwinskilukasz1, line 273: Isn't the first `ws*` superfluous? Otherwise it would also be needed in the first rule.  Also, let's align "return new" to each other.

czerwinskilukasz1 (changes requested):

mhagmajer:
@czerwinskilukasz1 what do you think about it?

src/askscript/parser/askscript.grammar.pegjs
czerwinskilukasz1: Well, it does not break for me:
image

image

and it actually shouldn't break, because `ws*` are consumed in mapEntryList, the only place where mapEntry is used:
image

czerwinskilukasz1:
> @czerwinskilukasz1 what do you think about it?

I think the change is great, just a nit in the grammar (see comment thread above) and it will be good to go.

src/askscript/parser/askscript.grammar.pegjs
YonatanKra: That is strange - it used to fail for me...
Maybe there was some caching issue with the grammer build locally.
Removed the `ws*`

mhagmajer (commented):
@czerwinskilukasz1 I'll leave it for you to accept and merge

czerwinskilukasz1 (approved):

Issue #379, opened 2020-07-09
Should we prevent swipe right from exiting Playground on Mac OS?


czerwinskilukasz1:
After I wrote quite a long program in AskScript in Playground, I accidentally swiped on my touchpad when trying to scroll the code window and left Playground. After I came back, the code was gone. 

How about we prevent swiping from doing this or ask for confirmation before leaving the page?
Alternatively we could remember the code in session or browser storage.

km4:
It's a great proposal! We can catch this event on the window if we change playground code (https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onbeforeunload)

if you have any other options please post it.

czerwinskilukasz1:
I was thinking of preventing swipe from going back to the previous webpage, but actually it would be great to detect going back or going forward for any reason and ask for confirmation. And if so, we should implement it for every OS.

czerwinskilukasz1:
Karol, would you like to work on it?
@km4

km4:
@czerwinskilukasz1 sure. Feel free to assign me.

km4:
@czerwinskilukasz1 I prefer to use `beforeunload` because not only back or forward detect but all events ( close window, refresh, change location). What do you think?

czerwinskilukasz1:
@km4 , ah, I must have missed your comment, I'm sorry!
Yes, `beforeunload` sounds very good then.

km4:
Ok, no problem, then I work on it. 👍

PR #515, opened 2020-10-13
feat(#379): confirm exit from  playground

km4:


src/playground/views/main.handlebars
```diff
@@ -149,6 +149,7 @@
 							<script>
 								const editor = registerAskScriptEditor("editor", "run", "{{askscriptServerUrl}}");
 								generateDemoMenuEntries(editor);
+								confirmExitFromPlayground(editor);
```
YonatanKra, line 152: Can you make it so this will work only if the editor is dirty?

km4: I can add this event on change in editor like editor.on('change', () => onbeforeunload ... )

YonatanKra: Sounds good.
Just need to make sure there's only one beforeunload event and not one per change event.
You could use your current code and maybe just during onchange add a dirty flag to the editor. Then check it in the beforeunload event?

mhagmajer: I like the @YonatanKra's suggestion

mhagmajer:
@czerwinskilukasz1 looks like @km4 is taking care of #379 that you have raised :)

src/playground/views/main.handlebars
km4: @YonatanKra @mhagmajer I think we can resolve this CR. Fill free to write another :P

src/playground/public/assets/js/app.ts
```diff
@@ -29,12 +29,18 @@ async function printAskVersion(elemId: string, askScriptServerUrl: string) {
 function registerAskScriptEditor(
   editorElementId: string,
   runElementId: string,
-  askScriptServerUrl: string
+  askScriptServerUrl: string,
+  dirty: boolean
```
YonatanKra, line 33: Have you thought of a use-case in which `dirty` will be `true` when you call this function?

km4: @YonatanKra no, I think that function should be initialized only once with default `false dirty`  when we start the playground. Should add some more? some conditions?

YonatanKra: Yea, move it inside. Much cleaner.
Also, remove the `change` listener once it is done.
Something like that:
```
const changeListenerCB = () => {
  confirmExitFromPlayground();
  editor.session.off('change', changeListenerCB);
}
editor.session.on('change', changeListenerCB);
```
Then you don't need the dirty flag at all. You just set the confirmation once a change happens and remove the change listener when you are done.
I'm not keen on the editor's API, so the syntax might vary.

mhagmajer (changes requested):
Thanks for this PR @km4! Please apply Yonatan's suggestion and we can go with this :+1:

src/playground/public/assets/js/app.ts
km4: @YonatanKra good point.
have some problem with `off`
Error code Property 'off' does not exist on type 'IEditSession'.
So I try to understand these problems in I found something like this https://github.com/thlorenz/brace/blob/master/index.d.ts#L458 . I can add `@ts-ignoge for this line or use `editor.off`
FYI: The `session.removeListener` throws the same error.

Maybe you have more experience with the ace editor.

km4:
@mhagmajer I see the light in the tunnel :P

src/playground/public/assets/js/app.ts
YonatanKra: Hi. Sorry. Didn't get the time to look at the editor api. Never used it before.
If you can't find a way to do it using the wditor api I'll take a look during the weekend.

mhagmajer, line 33: Looks like `dirty` should a variable local to this function rather than the argument

src/playground/public/assets/js/app.ts
```diff
@@ -107,3 +113,11 @@ function removeFadeInClass() {
   const resultElem = document.getElementById('result')!;
   resultElem.classList.remove('fadeIn');
 }
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
mhagmajer, line 117: this function is used so the comment shouldn't be needed

src/playground/public/assets/js/app.ts
```diff
@@ -29,12 +29,18 @@ async function printAskVersion(elemId: string, askScriptServerUrl: string) {
...
+  dirty: boolean
 ) {
   const editor = ace.edit(editorElementId);
   editor.setTheme('ace/theme/twilight');
   editor.session.setMode('ace/mode/javascript');
   editor.session.setOption('useWorker', false);
+  editor.session.on('change', function () {
+    if (dirty) return;
+    confirmExitFromPlayground();
+    dirty = true;
```
mhagmajer, line 42: There is no counterpart that would set `dirty` to false so we'll require confirmation from everyone once they have made any edit. Given that the purpose of the playground is to execute the inputed code maybe it would make sense to only require it if the code wasn't executed?

src/playground/public/assets/js/app.ts
```diff
@@ -107,3 +113,11 @@ function removeFadeInClass() {
   const resultElem = document.getElementById('result')!;
   resultElem.classList.remove('fadeIn');
 }
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+function confirmExitFromPlayground(): void {
+  window.addEventListener('beforeunload', function (e) {
+    e.preventDefault();
+    e.returnValue = '';
```
mhagmajer, line 121: I think it would make it easier to maintain if we alway attach this handler in `registerAskScriptEditor` but only execute these lines if `dirty` is `true`

mhagmajer (changes requested):
Thanks for doing this @km4 ! Just a few nits below

src/playground/public/assets/js/app.ts
km4: remove the flag from arguments

src/playground/public/assets/js/app.ts
km4: I stay with this solution, now the flag dirty change if we  run a script in the playground

src/playground/public/assets/js/app.ts
km4: add a condition for these handlers

src/playground/public/assets/js/app.ts
km4: you are right. remove it

km4:
@mhagmajer some update, check now and add feedback

src/playground/public/assets/js/app.ts
```diff
@@ -32,9 +32,15 @@ function registerAskScriptEditor(
...
   const editor = ace.edit(editorElementId);
+  let dirty = false;
   editor.setTheme('ace/theme/twilight');
   editor.session.setMode('ace/mode/javascript');
   editor.session.setOption('useWorker', false);
+  editor.session.on('change', () => {
+    if (dirty) return;
+    playgroundIsDirty(!dirty);
+    dirty = true;
+  });
```
mhagmajer, line 43: I think that now when we have `isDirty` check, we can just attach `onbeforeunload` once for all without the need for the extra `playgroundIsDirty` function.

```suggestion
  });
  
  window.onbeforeunload = function (e: any) {
    if (isDirty) {
      e.preventDefault();
      e.returnValue = '';
    }
  };
```

src/playground/public/assets/js/app.ts
```diff
@@ -81,6 +87,7 @@ async function executeAskScript(
     const json = await response.json();
     if (response.status == 200) {
       showSuccessfulResponse(json.result);
+      playgroundIsDirty(false);
```
mhagmajer, line 90:
```suggestion
      isDirty = false;
```

mhagmajer (changes requested):
Requesting some finishing touches :)

src/playground/public/assets/js/app.ts
km4: what about this isDirty flag? Now is undefined variable for this scope. If I push dirty params to all functions eg. 
```
async function executeAskScript(
  askScriptCode: string,
  askScriptServerUrl: string,
  isDirty: boolean
) {
```
`executeAskScriptFromEditor(editor, askScriptServerUrl, isDirty); ` the flag dosnt update in the top.

PR #523, opened 2020-10-22
fix(at): throw error when using at to access undefined property

### LINKED ISSUE #404: at should throw if the key is missing
```
ask {
    const marta = { firstName: 'Marta' }
    marta:at('firstName3')
}
```
returns `null` (empty) but an error was expected (playground with AskVM v. 1.2.0)

dastin-sandura:
closes #404

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "askql",
-  "version": "1.2.1",
+  "version": "1.2.2",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/src/askvm/resources/core/at.ts
+++ b/src/askvm/resources/core/at.ts
@@ -9,11 +9,7 @@ export const at = resource({
       (listOrObject.length <= key || key < 0)
     )
       throw Error(
-        'Sorry, but index ' +
-          key +
-          ' is out of bounds for a list with ' +
-          listOrObject.length +
-          ' elements.'
+        `Sorry, but index ${key} is out of bounds for a list with ${listOrObject.length} elements.`
       );
 
     if (
--- a/src/askvm/resources/core/at.ts
+++ b/src/askvm/resources/core/at.ts
@@ -5,7 +5,7 @@ export const at = resource({
   async resolver(listOrObject: any, key: any): Promise<any> {
     if (!(key in listOrObject))
       throw Error(
-        'Requested property key was not defined so it is not possible to get its value.NO'
+        'Requested property key was not defined so it is not possible to get its value.'
       );
     return typed(listOrObject[key]);
   },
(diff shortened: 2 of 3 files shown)


mhagmajer:
@dastin-sandura thanks for creating a PR, it looks nice! You have covered the object case we'll, buy what about the case where `at` is used on a list? Can we cover that as well and add a relevant unit test?

```
const list =[1,2,3]
list[3]
```
Should probably also throw. Note that the code above is in AskScript and after transpilation to AskCode it will use the `at` resource.

dastin-sandura:
Sure @mhagmajer :) I will add a test and implementation for your case.
As I understand your requirement, It will work something like '[index out of bounds exception](https://docs.oracle.com/javase/7/docs/api/java/lang/IndexOutOfBoundsException.html)' in Java?

mhagmajer:
@dastin-sandura yes, that would be great and consistent with the behavior you have already implemented 👍

dastin-sandura:
I will add it and let you know when done 😃

dastin-sandura:
@mhagmajer done and ready for review 😄

YonatanKra (approved):
LGTM - @mhagmajer ?

dastin-sandura:
@mhagmajer could you make this pull request count towards Hacktoberfest 2020? (https://hacktoberfest.digitalocean.com/)

dastin-sandura:
@mhagmajer I checked my Hacktoberfest account and I see that it already counts 👍

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "askql",
-  "version": "1.2.1",
+  "version": "1.2.2",
```
mhagmajer, line 3: I forgot to update in on the last release :+1:

src/askvm/resources/core/at.ts
```diff
@@ -3,6 +3,27 @@ import { any, resource, typed } from '../../lib';
...
+      Array.isArray(listOrObject) &&
+      Number.isInteger(key) &&
+      (listOrObject.length <= key || key < 0)
+    )
+      throw Error(
+        'Sorry, but index ' +
+          key +
+          ' is out of bounds for a list with ' +
+          listOrObject.length +
+          ' elements.'
```
mhagmajer, line 16: Can we use [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) here for greater visibility?

mhagmajer (changes requested):
Just a nit to fix :)

src/askvm/resources/core/at.ts
dastin-sandura: sure 🙂

dastin-sandura: @mhagmajer btw thanks for the direct link to documentation about template literals 😄

dastin-sandura (commented):
Refactored using template literals 👍

mhagmajer (approved):
Thanks!

PR #524, opened 2020-10-24
feat(playground): added more playground examples

### LINKED ISSUE #181: Add more examples to the AskScript Playground

YonatanKra:
Closes #181

### DIFF
--- a/src/playground-backend/lib/resources.ts
+++ b/src/playground-backend/lib/resources.ts
@@ -1,4 +1,5 @@
 import { Resources, resource } from '../../askvm';
+import fetch from 'node-fetch';
 
 export const customResources: Resources = {
   withSmile: resource<string, [string]>({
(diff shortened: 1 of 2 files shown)


src/playground/public/assets/js/demosData.ts
```diff
@@ -108,4 +108,44 @@ ask {
...
+    code: `ask {
+      fetch('https://api.covid19india.org/data.json')['cases_time_series']
+      :map(fun(dataSet) {
+                             return {
+                                data: dataSet['date'],
+                                dailyconfirmed: dataSet['dailyconfirmed'],
+                                dailydeceased: dataSet['dailydeceased'],
+                                dailyrecovered: dataSet['dailyrecovered']
+                              }
+                            })
```
mhagmajer, line 123: let's fix the indentation

mhagmajer: Also `return is not needed here

src/playground-backend/lib/resources.ts
```diff
@@ -12,4 +13,32 @@ export const customResources: Resources = {
       return "Hi, this is AskQL server! It's " + new Date().toString();
     },
   }),
+  statewise: resource({
```
mhagmajer, line 16: Let's make this resource name more specific

src/playground/public/assets/js/demosData.ts
mhagmajer: This mapping is essentially picking keys from objects. Shall we change key names to make this a little more exciting?

mhagmajer (changes requested):
Thanks for adding more examples @YonatanKra.

All of those are currently using `fetch` and `map` maybe we can use add `filter` to some of them to make them more interesting?

Issue #521, opened 2020-10-18
Refactor let


YonatanKra:
Currently let holds the logic for `assign` and `const` which are different concerns.
Should write tests and implement `const` and `assign`.

dastin-sandura:
Hey @YonatanKra can I take care of this issue? :)

YonatanKra:
@dastin-sandura - the idea would be to split these into their own files.
Currently all of these logics are the same.

You can see it in this line:
`export { letRes as let, letRes as const, letRes as assign } from './let';`

Thing is, they are now quite different than each other. Enough to see the `if` statements inside the file like:
`if (code.name === 'assign') {`
`if (code.name === 'const') {`

So this task is a refactor task:

Split the `let` file into 3 files (const, assign, let) and extract a common module if needed.

They all should be covered with tests as well (preferably before the process begins). I've already added some tests in the `let.test.ts` file but since I "hopped in" just to add a feature, but note they are mixed (some test let, some test assign, some test const). They might not cover all of the use cases for the 3 resources.

dastin-sandura:
Thank you! @YonatanKra 😄 this clears things up! I did not notice the ```export``` detail when I was analyzing the code so this explains a lot. I will submit changes today in the evening CET.

PR #541, opened 2020-11-04
Issue 530  refactor reserved keyword logic

### LINKED ISSUE #530: Refactor: extract common logic for checking if a reserved keyword is used in from let, const and assign to avoid code duplication
**Is your feature request related to a problem? If yes, please describe the problem.**
It is not a problem, because the code works well and it is not visible to the user. It can be considered a 'code smell' by some devs.

**Describe the solution you'd like to see.**
Removed duplication of code 😄 

**Were you inspired by anything else, that you have experienced?**
Inspired by @YonatanKra 

**Describe alternatives you've considered (if any). If you haven't that is also ok :smile:**
The alternative is leaving the duplicated code, but as new people contribute this will become more and more problematic.

**Any additional details about this request that you want to share?**
Nope

**How important is it for you to have this feature?**
It could save time in the future because if there is only one implementation, changes will be made in one place instead of 3.

dastin-sandura:
closes #530


dastin-sandura:
Hey @YonatanKra , this is a continuation of refactoring from 'Refactor let #521' 😄

Issue #580, opened 2021-04-23
Let developers to call JS methods on limited JS objects 


pkarw:
**Is your feature request related to a problem? If yes, please describe the problem.**

It would be great if we could pass over some libraries - take the `lodash` to AskQL and let the developers use all the functions provided, without the need to pack every call with a painstakingly created resource (current case). It could be pretty much useful for using the external services APIs like [`commercetools sdk`](https://www.npmjs.com/package/@commercetools/typescript-sdk) or ORMs generated from tools like [`Prisma`](https://www.prisma.io/)

To keep it safe we can let call just the top-level functions, or add whitelisting + only on the explicitly passed resources like:

```js
const vmContext = { 
customResources:  [ 
 "lodash": jsModuleInterop(_, ["sort", "unique"])
]

```
where the `jsModuleInterop` is a wrapper over the JS object accepting the object (in this case `_` - lodash module, plus a whitelist of methods)

czerwinskilukasz1:
Sounds good. In case of lodash probably there are no harmful functions, but in general it would be good to have `jsModuleInterop`.

pkarw:
Cool! Any insights on how can we implement this feature? I was looking at the `run` method inside the `vm` implementation. Is it the right hook-up point to extend it? I mean we could potentially implement a "hack": to check the 1st parameter of any function call (as it should be the processing resource): if there is a proper child function - lets' call it.  However, it's not an ideal one. I mean when these functions are going to return JS objects (chaining) we'd be not able to use it further.

czerwinskilukasz1:
I envision `jsModuleInterop` the following way:
```js
const vmContext = { 
  customResources: [ 
   'lodash': jsModuleInterop('lodash', ['sort', 'unique'])` 
  ]
...
}
```

What `jsModuleInterop(arg1, args)` would be is a helper that creates a resource which has properties listed in `args` that are wrappers over functions with the same names from module `arg1`.

czerwinskilukasz1:
This way we don't need to add a hack with additional checks for function names in runtime. We use an existing mechanism instead.

pkarw:
That's cool, however I've got an additional question. What do you mean by resource properties? I mean at the moment one resource could have just a single `execute`  and secondly `resolve` method - so no additional properties could be set. Or am I wrong?

pkarw:
@czerwinskilukasz1 maybe another option would be to add a support for something like JS scripting inside AskQL? I mean something like:

```
js { 
}
```

[...]

PR #590, opened 2021-05-20
feat(#580): a basic implementation of js function calling from askcript

pkarw:
With this feature developers get a new helper/factory method which can be used for passing JS modules and functions to AskScript without the need of manually warping them with resources.

Usage:

```js
import { factory } from 'askql/askvm/resource';
const moduleToBeWrapped = {
  wrapMe1: () => {
    return 100;
  },
  wrapMe2: (arg: string) => {
    return arg;
  },
  dontWrapMe: () => {
    return 0;
  },
};

function ask(code: string) {
  return runUntyped(
    {
      resources: {  ...factory(moduleToBeWrapped, ['wrapMe1', 'wrapMe2'])},
      values,
    },
    parse(code)
  );
}

// now you can call the function `wrapMe1` of `moduleToBeWrapped` in AskScript
ask(`ask(call(get('wrapMe1')))

```

You can easily wrap a whole module like `lodash` by just simply:

```js
  return runUntyped(
    {
      resources: {  ...factory(lodash, ['sort'])},
      values,
    },
    parse(code)
  );
```

### DIFF
--- a/src/askvm/lib/resource.ts
+++ b/src/askvm/lib/resource.ts
@@ -46,4 +46,26 @@ export function resource<T, A extends any[]>(
   return new Resource(...options);
 }
 
+/**
+ * Resource factory generates the resource wrappers around all the methods which are on `whitelist` of the `module` passed. Implementation of https://github.com/CatchTheTornado/askql/issues/580 for limited JS function calls inside AskScript
+ * @param module JS object which methods will be wrapped as resource handlers
+ * @param whitelist list of strings including function/method names of the module to be wrapped out
+ */
+export function factory(
+  jsModule: any,
+  whitelist: string[]
+): Record<string, Resource<any, any>> {
+  const res: Record<string, Resource<any, any>> = {};
+  Object.keys(jsModule).forEach((k) => {
+    if (whitelist.includes(k)) {
+      res[k] = resource({
+        type: any,
+        async resolver(...args: any) {
+          return jsModule[k](...args);
+        },
+      });
+    }
+  });
+  return res;
+}
 export type Resources = Record<string, Resource<any, any>>;
--- a/src/askvm/resources/core/__tests__/factory.test.ts
+++ b/src/askvm/resources/core/__tests__/factory.test.ts
@@ -1,0 +1,48 @@
+import { runUntyped } from '../../../index';
+import { parse } from '../../../../askcode/lib';
+import { factory } from '../../../lib/resource';
+import * as core from '..';
+
+const values = {};
+const moduleToBeWrapped = {
+  wrapMe1: () => {
+    return 100;
+  },
+  wrapMe2: (arg: string) => {
+    return arg;
+  },
+  dontWrapMe: () => {
+    return 0;
+  },
+};
+const wrappedResources = factory(moduleToBeWrapped, ['wrapMe1', 'wrapMe2']);
+
+function ask(code: string) {
+  return runUntyped(
+    {
+      resources: { ...core, ...wrappedResources },
+      values,
+    },
+    parse(code)
+  );
+}
+
+describe(`factory`, function () {
+  it(`factory should wrap only whitelisted resources`, async function () {
+    await expect(
+      Promise.resolve(Object.keys(wrappedResources))
+    ).resolves.toStrictEqual(['wrapMe1', 'wrapMe2']);
+  });
+  it(`should wrap wrapMe1 js method and return 100`, async function () {
+    const expectedResult = 100;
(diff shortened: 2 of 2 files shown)


lukasz-xfaang (approved):
Looks good to me

czerwinskilukasz1 (approved):
Looks good to me

src/askvm/resources/core/__tests__/factory.test.ts
```diff
@@ -0,0 +1,48 @@
...
+      Promise.resolve(Object.keys(wrappedResources))
+    ).resolves.toStrictEqual(['wrapMe1', 'wrapMe2']);
+  });
+  it(`should wrap wrapMe1 js method and return 100`, async function () {
+    const expectedResult = 100;
+    await expect(ask(`ask(call(get('wrapMe1')))`)).resolves.toStrictEqual(
+      expectedResult
+    );
+  });
+  it(`should wrap wrapMe2 js method nad return arg`, async function () {
```
czerwinskilukasz1, line 42: Typo: 'nad' -> 'and'

src/askvm/resources/core/__tests__/factory.test.ts
```diff
@@ -0,0 +1,48 @@
...
+    const expectedResult = 'arg';
```
czerwinskilukasz1, line 43: Maybe using a different string than 'arg' would be better, so that it's not the same as the variable name.

src/askvm/lib/resource.ts
```diff
@@ -46,4 +46,26 @@ export function resource<T, A extends any[]>(
...
+ * @param module JS object which methods will be wrapped as resource handlers
+ * @param whitelist list of strings including function/method names of the module to be wrapped out
+ */
+export function factory(
+  jsModule: any,
+  whitelist: string[]
+): Record<string, Resource<any, any>> {
+  const res: Record<string, Resource<any, any>> = {};
+  Object.keys(jsModule).forEach((k) => {
+    if (whitelist.includes(k)) {
```
czerwinskilukasz1, line 60: I am wondering how big the library could be and how suboptimal it would be if there were thousands of keys in `jsModule`.  In the future, we might want to iterate rather through `whitelist` instead, but premature optimization is evil so let's keep it like this for now.

czerwinskilukasz1:
@mhagmajer , do you have any specific comments on this PR?

@pkarw , I added a few minor comments if you had time, but feel free to merge it without addressing them.

src/askvm/lib/resource.ts
```diff
@@ -46,4 +46,26 @@ export function resource<T, A extends any[]>(
   return new Resource(...options);
 }
 
+/**
+ * Resource factory generates the resource wrappers around all the methods which are on `whitelist` of the `module` passed. Implementation of https://github.com/CatchTheTornado/askql/issues/580 for limited JS function calls inside AskScript
+ * @param module JS object which methods will be wrapped as resource handlers
+ * @param whitelist list of strings including function/method names of the module to be wrapped out
+ */
+export function factory(
```
mhagmajer, line 54: One function per file please

src/askvm/resources/core/__tests__/factory.test.ts
```diff
@@ -0,0 +1,48 @@
...
+    await expect(
```
mhagmajer, line 44: Maybe we can also check if the wrapped function was given the right arguments if more than one

src/askvm/resources/core/__tests__/factory.test.ts
```diff
@@ -0,0 +1,48 @@
...
+      ask(`ask(call(get('wrapMe2'), 'arg'))`)
```
mhagmajer, line 45: We could also check if it works with asynchronous functions

src/askvm/resources/core/__tests__/factory.test.ts
pkarw: Fixed

src/askvm/lib/resource.ts
pkarw: Fixed

src/askvm/lib/resource.ts
pkarw: OK, fixed

src/askvm/resources/core/__tests__/factory.test.ts
pkarw: I suppose it's not specific to this resource (I mean `async` behavior) I'm gonna skip it :)

PR #591, opened 2021-05-27
chore: prevent Dependabot from creating PRs for patch updates

### LINKED ISSUE #214: Prevent dependabot from upgrading patch versions
A number of PRs from dependabot regard a patch update, meaning from A.B.X to A.B.Y, which is a minor fix, e.g. here: https://github.com/xFAANG/askql/pull/207
It would be great if it created PRs only for major and minor updates in version number (A and B).

czerwinskilukasz1:
(Hopefully) closes #214

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,3 +7,11 @@ updates:
     open-pull-requests-limit: 10
     labels:
       - enhancement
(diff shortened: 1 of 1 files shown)


mhagmajer (approved):
Sounds good!

